PR #58, opened 2023-09-04
Initialisation strategy

vahidlazio:
- Add the option to choose an initialisation strategy 
- ActivateAndFetchAsync or FetchAndActivate are the different strategies that can be selected in the initialisation phase

### DIFF
--- a/Provider/src/main/java/dev/openfeature/contrib/providers/ConfidenceFeatureProvider.kt
+++ b/Provider/src/main/java/dev/openfeature/contrib/providers/ConfidenceFeatureProvider.kt
@@ -3,6 +3,8 @@ package dev.openfeature.contrib.providers
 import android.content.Context
 import dev.openfeature.contrib.providers.apply.FlagApplier
 import dev.openfeature.contrib.providers.apply.FlagApplierWithRetries
+import dev.openfeature.contrib.providers.cache.DiskStorage
+import dev.openfeature.contrib.providers.cache.InMemoryCache
 import dev.openfeature.contrib.providers.cache.ProviderCache
 import dev.openfeature.contrib.providers.cache.ProviderCache.CacheResolveResult
 import dev.openfeature.contrib.providers.cache.StorageFileCache
@@ -41,6 +43,8 @@ class ConfidenceFeatureProvider private constructor(
     override val hooks: List<Hook<*>>,
     override val metadata: ProviderMetadata,
     private val cache: ProviderCache,
+    private val storage: DiskStorage,
+    private val initialisationStrategy: InitialisationStrategy,
     private val client: ConfidenceClient,
     private val flagApplier: FlagApplier,
     private val eventsPublisher: EventsPublisher,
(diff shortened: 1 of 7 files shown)


Provider/src/main/java/dev/openfeature/contrib/providers/ConfidenceFeatureProvider.kt
```diff
@@ -54,14 +58,43 @@ class ConfidenceFeatureProvider private constructor(
             eventsPublisher.publish(OpenFeatureEvents.ProviderReady)
```
nicklasl, line 58: Maybe we should to send `ProviderError` if the cache is empty at this point? Or `ProviderStale`?

Provider/src/main/java/dev/openfeature/contrib/providers/ConfidenceFeatureProvider.kt
```diff
@@ -54,14 +58,43 @@ class ConfidenceFeatureProvider private constructor(
...
+
     override fun initialize(initialContext: EvaluationContext?) {
         if (initialContext == null) return
+
+        // refresh cache with the last stored data
+        storage.read()?.let(cache::refresh)
+
+        if (initialisationStrategy == InitialisationStrategy.ActivateAndFetchAsync) {
+            eventsPublisher.publish(OpenFeatureEvents.ProviderReady)
+        }
```
nicklasl, line 70: Like this a lot, makes things very readable!

Provider/src/main/java/dev/openfeature/contrib/providers/ConfidenceFeatureProvider.kt
```diff
@@ -184,15 +217,25 @@ class ConfidenceFeatureProvider private constructor(
...
+
         @Suppress("LongParameterList")
         fun create(
             context: Context,
             clientSecret: String,
             region: ConfidenceRegion = ConfidenceRegion.EUROPE,
+            initialisationStrategy: InitialisationStrategy = InitialisationStrategy.FetchAndActivate,
             hooks: List<Hook<*>> = listOf(),
             client: ConfidenceClient? = null,
             metadata: ProviderMetadata = ConfidenceMetadata(),
```
nicklasl, line 236: For another time: We should probably align on when we use default params. In this `create` function or in the constructor.

Provider/src/main/java/dev/openfeature/contrib/providers/ConfidenceFeatureProvider.kt
```diff
@@ -54,14 +58,43 @@ class ConfidenceFeatureProvider private constructor(
             eventsPublisher.publish(OpenFeatureEvents.ProviderReady)
         }
     }
+
     override fun initialize(initialContext: EvaluationContext?) {
         if (initialContext == null) return
+
+        // refresh cache with the last stored data
+        storage.read()?.let(cache::refresh)
```
nickybondarenko, line 66: Shall we add some error handling here in case storage data is corrupted?

Provider/src/main/java/dev/openfeature/contrib/providers/cache/StorageFileCache.kt
```diff
@@ -3,48 +3,84 @@ package dev.openfeature.contrib.providers.cache
...
+import dev.openfeature.sdk.Value
+import kotlinx.serialization.Serializable
 import kotlinx.serialization.encodeToString
 import kotlinx.serialization.json.Json
 import java.io.File
 
 const val FLAGS_FILE_NAME = "confidence_flags_cache.json"
 
-class StorageFileCache private constructor(context: Context) : InMemoryCache() {
+interface DiskStorage {
```
nickybondarenko, line 14: The file should also be renamed 👍

Provider/src/main/java/dev/openfeature/contrib/providers/ConfidenceFeatureProvider.kt
nicklasl: Devils advocate: What can we do? And how can we detect a corrupt storage data?

nickybondarenko: If we're not able to read the storage for some reason when it's not null, we should handle that - perhaps by wiping storage?

vahidlazio: not being able to read the storage is typically an OS error and it's rare but technically can happen. I believe wiping the data wouldn't solve the problem that OS failed to read the data. wdyt?

nickybondarenko: Makes sense to me! Will mark as resolved 👍

Provider/src/main/java/dev/openfeature/contrib/providers/ConfidenceFeatureProvider.kt
vahidlazio: this exception happens when the network request has failed. especially when the initialisation strategy is activate and fetch IMO it doesn't make sense to have provider not ready. for fetch and activate, maybe .. but is it what we want? considering that if the developer decides to have the strategy as "try to refresh first" or "force refresh first"? in the latter case they can't use the properties since the provider is never ready. they will be stuck. but if the definition of FetchAndActivate is try to fetch, then the failure in the network request should be treated as ready? we have to add this to the documentation anyways.

nickybondarenko: I agree, with ActivateAndFetch, the provider should be ready, since the developer would most likely apply the flags that were fetched later in the lifecycle / next session.
For fetchAndActivate - would we then handle an error as "nothing new to fetch" and wait for another opportunity to fetch freshest flags? In that case, I also agree that it can be a ProviderReady. My only concern (which might come from me not fully understanding Confidence/data science) is whether we report these events correctly.

nickybondarenko (approved):

PR #67, opened 2023-09-06
move the event handlings to the provider instead of using the static instance

vahidlazio:


nicklasl (approved):
Good stuff!
I also opened an issue with the SDK where to remove the singleton pattern.

nicklasl:
> Good stuff! I also opened an issue with the SDK where to remove the singleton pattern.

When starting to do this I was think I should also check the [spec about events](https://github.com/open-feature/spec/blob/main/specification/sections/05-events.md) and that is written (seems to me) with the intent of the Events being "global" to the SDK which would then mean that if the SDK runs with multiple providers (something we don't support yet? - but really what is happening when we run tests in parallel), each event should really echo which provider it came from?

vahidlazio:
@nicklasl if we support multiple providers i don't see another way as stating in the events the provider name or hash or some sort of identifier, means the providers should also have an identifier in the spec. which i believe having the events coming from the providers and hosted and controlled by the providers is a way better setup.

nicklasl:
Will using this function on the OpenFeature API work with this implementation?
```
OpenFeatureAPI.observeEvents
````

**Comment by vahidlazio** [comment]
@nicklasl 
```
Will using this function on the OpenFeature API work with this implementation?

```

I believe It does however to support multiple providers we should think of the identifier and pass it to the event observer function as well. right?

File: Provider/src/main/java/dev/openfeature/contrib/providers/ConfidenceFeatureProvider.kt [hunk 1]
```diff
@@ -253,4 +272,38 @@ class ConfidenceFeatureProvider private constructor(
...
+    coroutineScope.launch {
+        observeProviderReady()
+            .take(1)
+            .collect {
+                continuation.resumeWith(Result.success(Unit))
+            }
+    }
+
+    coroutineScope.launch {
+        EventHandler.eventsObserver()
```
nicklasl, line 290: wouldn't this use the global (singleton) eventshandler instead of the one created in the `provider`?

PR #69, opened 2023-09-11
fix: On context change

vahidlazio:
* Stale event when the context is changed
* Provider Ready event when the fetch is completed after a context change
* new values are updated in the cache after a successful fetch following a context change.

### DIFF
--- a/Provider/src/main/java/dev/openfeature/contrib/providers/ConfidenceFeatureProvider.kt
+++ b/Provider/src/main/java/dev/openfeature/contrib/providers/ConfidenceFeatureProvider.kt
@@ -60,12 +60,22 @@ class ConfidenceFeatureProvider private constructor(
     }
 
     override fun initialize(initialContext: EvaluationContext?) {
-        if (initialContext == null) return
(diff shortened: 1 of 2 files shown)


Provider/src/main/java/dev/openfeature/contrib/providers/ConfidenceFeatureProvider.kt
```diff
@@ -108,7 +118,14 @@ class ConfidenceFeatureProvider private constructor(
         newContext: EvaluationContext
     ) {
         if (newContext != oldContext) {
-            initialize(newContext)
+            eventsPublisher.publish(OpenFeatureEvents.ProviderStale)
+
+            // on the new context
+            // we want to fetch new values and update te storage & cache
```
nicklasl, line 124:
```suggestion
            // on the new context we want to fetch new values and update
            // the storage & cache right away which is why we pass `InitialisationStrategy.FetchAndActivate`
```

nicklasl (approved):
LGTM

PR #78, opened 2023-10-04
ci: publish to maven central

nicklasl:
When running the release-please pipeline the CI steps will:
- import a base64 encoded gpg key defined in a secret variable
- export that key to a `secring.gpg` file
- construct a `gradle.properties` file that contains `signing.keyId`, `signing.password` & `signing.secretKeyRingFile`. These are properties that gradle needs to sign the artifacts.
- run `./gradlew :Provider:publish` which will use env variables and gradle properties to sign and publish the artfacts to sonatype nexus.


Provider/build.gradle.kts
```diff
@@ -60,31 +68,30 @@ dependencies {
...
-        publications {
-            register<MavenPublication>("release") {
-                groupId = "com.spotify"
-                artifactId = "confidence-openfeature-provider-android"
-                version = Versions.providerVersion
+publishing {
+    publications {
+        register<MavenPublication>("release") {
+            groupId = "com.spotify"
+            artifactId = "confidence-openfeature-provider-android"
```
nicklasl, line 75:
```suggestion
            groupId = "com.spotify.confidence"
            artifactId = "openfeature-provider-android"
```

nicklasl: @fabriziodemaria what do you think?

nickybondarenko:
LGTM, I think the groupId change should be added though!

PR #80, opened 2023-10-06
fix: Move apply cache into disk storage

nicklasl:
All file i/o should now be contained behind the DiskStorage interface which will improve testability.

### DIFF
--- a/Provider/src/main/java/com/spotify/confidence/cache/StorageFileCache.kt
+++ b/Provider/src/main/java/com/spotify/confidence/cache/StorageFileCache.kt
@@ -14,10 +14,10 @@ import kotlinx.serialization.modules.SerializersModule
 import kotlinx.serialization.modules.contextual
 import java.io.File
 
-const val FLAGS_FILE_NAME = "confidence_flags_cache.json"
-const val APPLY_FILE_NAME = "confidence_apply_cache.json"
+internal const val FLAGS_FILE_NAME = "confidence_flags_cache.json"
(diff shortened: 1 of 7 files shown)


Provider/src/main/java/com/spotify/confidence/cache/StorageFileCache.kt
```diff
@@ -1,18 +1,26 @@
...
+import kotlinx.serialization.modules.SerializersModule
+import kotlinx.serialization.modules.contextual
 import java.io.File
 
 const val FLAGS_FILE_NAME = "confidence_flags_cache.json"
+const val APPLY_FILE_NAME = "confidence_apply_cache.json"
 
-class StorageFileCache private constructor(context: Context) : DiskStorage {
-    private val file: File = File(context.filesDir, FLAGS_FILE_NAME)
+class StorageFileCache private constructor(
```
vahidlazio, line 20: great stuff! can we go one step further and make this private/internal?

nicklasl: Good point. fixed!

vahidlazio (dismissed):

Provider/src/main/java/com/spotify/confidence/cache/StorageFileCache.kt
nicklasl: We could probably go through and do this for more classes, maybe not the interfaces but definitely the implementations.

vahidlazio (approved):

PR #81, opened 2023-10-09
fix: add pom declaration to maven publication

nickybondarenko:
As a part of publishing to Maven Central, this PR adds declaration of licensing, developers, project links and source control management (scm).

### DIFF
--- a/Provider/build.gradle.kts
+++ b/Provider/build.gradle.kts
@@ -85,6 +85,48 @@ publishing {
                     }
                 }
             }
+            pom {
+                name.set("Confidence Openfeature Provider Android")
+                description.set("An Openfeature Provider for Confidence, made for the Android SDK")
+                url.set("https://github.com/spotify/confidence-openfeature-provider-kotlin")
+                licenses {
+                    license {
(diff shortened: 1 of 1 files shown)


Provider/build.gradle.kts
```diff
@@ -85,6 +85,48 @@ publishing {
                     }
                 }
             }
+            pom {
+                name.set("Confidence SDK Provider Android")
+                description.set("Confidence SDK Provider Android")
```
nicklasl, line 90: Drop the `SDK`.
```suggestion
                name.set("Confidence Openfeature Provider Android")
                description.set("An Openfeature Provider for Confidence, made for the Android SDK")
```

Provider/build.gradle.kts
```diff
@@ -85,6 +85,48 @@ publishing {
...
+                url.set("https://confidence.spotify.com/")
```
nicklasl, line 91: in the [java pom](https://github.com/spotify/confidence-openfeature-provider-java/blob/main/pom.xml) we use the github url here... not at all sure what is more correct :)

PR #102, opened 2023-11-13
feat: Add SDK id and version to requests

fabriziodemaria:
To comply with the updated Confidence APIs.

### DIFF
--- a/Provider/src/main/java/com/spotify/confidence/ConfidenceFeatureProvider.kt
+++ b/Provider/src/main/java/com/spotify/confidence/ConfidenceFeatureProvider.kt
@@ -12,6 +12,7 @@ import com.spotify.confidence.client.ConfidenceClient
 import com.spotify.confidence.client.ConfidenceRegion
 import com.spotify.confidence.client.ConfidenceRemoteClient
 import com.spotify.confidence.client.ResolveResponse
+import com.spotify.confidence.client.SdkMetadata
 import dev.openfeature.sdk.EvaluationContext
 import dev.openfeature.sdk.FeatureProvider
 import dev.openfeature.sdk.Hook
@@ -33,6 +34,8 @@ import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.SupervisorJob
 import kotlinx.coroutines.cancelChildren
 import kotlinx.coroutines.launch
+
+const val SDK_ID = "SDK_ID_KOTLIN_PROVIDER"
 
 @Suppress(
     "TooManyFunctions",
@@ -258,6 +261,7 @@ class ConfidenceFeatureProvider private constructor(
         ): ConfidenceFeatureProvider {
             val configuredClient = client ?: ConfidenceRemoteClient(
                 clientSecret = clientSecret,
+                sdkMetadata = SdkMetadata(SDK_ID, BuildConfig.SDK_VERSION),
                 region = region,
                 dispatcher = dispatcher
             )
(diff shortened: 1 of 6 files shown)


Provider/src/main/java/com/spotify/confidence/ConfidenceFeatureProvider.kt
```diff
@@ -258,6 +258,10 @@ class ConfidenceFeatureProvider private constructor(
         ): ConfidenceFeatureProvider {
             val configuredClient = client ?: ConfidenceRemoteClient(
                 clientSecret = clientSecret,
+                sdkId = "SDK_ID_KOTLIN_PROVIDER",
```
fabriziodemaria, line 261: Discussions are happening to change the OF SDK naming from `kotlin` to `android`. Shall we follow the same route, we might want to rename this

release-please-config.json
```diff
@@ -10,6 +10,7 @@
             "extra-files": [
                 "README.md",
                 "Provider/build.gradle.kts"
+                "Provider/src/main/java/com/spotify/confidence.ConfidenceFeatureProvider.kts"
```
fabriziodemaria, line 13: Would this even work? And how bad practice is this?
It's nice to imagine a solution based off of release-please that is easy and works the same way in all providers, but this method would have a few important back-draws

Provider/src/main/java/com/spotify/confidence/ConfidenceFeatureProvider.kt
```diff
@@ -258,6 +258,10 @@ class ConfidenceFeatureProvider private constructor(
...
+                // x-release-please-start-version
+                sdkVersion = "0.1.2",
+                // x-release-please-end
```
nicklasl, line 264: This was an interesting approach but I would suggest we use `BuildConfig` instead. 
You can see an example how it is defined [here](https://github.com/spotify/confidence-openfeature-provider-kotlin/blob/main/ConfidenceDemoApp/build.gradle.kts#L41) and used [here](https://github.com/spotify/confidence-openfeature-provider-kotlin/blob/8a1ece8ba4aff7a189f4d10cc38fd279d1d67b27/ConfidenceDemoApp/src/main/java/com/example/confidencedemoapp/ClientSecretProvider.kt#L4).

Provider/src/main/java/com/spotify/confidence/client/ConfidenceRemoteClient.kt
```diff
@@ -20,6 +20,8 @@ import java.net.HttpURLConnection
 
 class ConfidenceRemoteClient : ConfidenceClient {
     private val clientSecret: String
+    private val sdkId: String
```
vahidlazio, line 23: if we add these 2 to another object (creating a new type), then we can add more stuff to it in the future if needed easier.

Provider/src/main/java/com/spotify/confidence/ConfidenceFeatureProvider.kt
Billlynch: you can also use the inline comment `// x-release-please-version` too

Provider/src/main/java/com/spotify/confidence/ConfidenceFeatureProvider.kt
fabriziodemaria: A problem for another PR

Provider/src/main/java/com/spotify/confidence/ConfidenceFeatureProvider.kt
```diff
@@ -258,6 +261,7 @@ class ConfidenceFeatureProvider private constructor(
         ): ConfidenceFeatureProvider {
             val configuredClient = client ?: ConfidenceRemoteClient(
                 clientSecret = clientSecret,
+                sdkMetadata = SdkMetadata(SDK_ID, BuildConfig.SDK_VERSION),
```
fabriziodemaria, line 264: Doesn't work, the result string for version is: `Build_gradle$Versions@1e4d110c.providerVersion`. To be fixed

nicklasl (approved):

PR #107, opened 2023-12-15
chore: do not expose the artefact for open feature sdk

vahidlazio:


nicklasl:
What do we/or the consumers of this SDK gain from this change?

vahidlazio:
@nicklasl ideally the consumers have one kotlin sdk and they can switch to different providers or have multiple providers, if the providers expose their own sdk this will be problematic in the future.

nicklasl:
> @nicklasl ideally the consumers have one kotlin sdk and they can switch to different providers or have multiple providers, if the providers expose their own sdk this will be problematic in the future.

OK, I can get on board that. But we would need to update the `README.md` so it makes sense + keep that up to date with any SDK bumps we do.

PR #115, opened 2024-01-08
chore: add a default endpoint to the provider

nickybondarenko:
Previously, the region would default to Europe. This PR is changing this to GLOBAL and is specifying a URL for that.


Provider/src/main/java/com/spotify/confidence/client/ConfidenceRemoteClient.kt
```diff
@@ -45,6 +45,7 @@ class ConfidenceRemoteClient : ConfidenceClient {
             "application/json"
         )
         baseUrl = when (region) {
+            ConfidenceRegion.GLOBAL -> "https://resolver.confidence.dev/"
```
nicklasl, line 48:
```suggestion
            ConfidenceRegion.GLOBAL -> "https://resolver.confidence.dev"
```
please drop the last `/` to be consistent.

nicklasl (approved):

PR #117, opened 2024-01-15
fix: Fix sdk metadata format

fabriziodemaria:


Provider/src/main/java/com/spotify/confidence/client/Types.kt
```diff
@@ -20,8 +20,14 @@ data class ResolveFlagsRequest(
...
     val clientSecret: String,
     val apply: Boolean,
-    val sdkId: String,
-    val sdkVersion: String
+    val sdk: Sdk
+)
+
+@Serializable
+data class Sdk(
+    @Contextual
```
vahidlazio, line 28: I think we don't need this and it can be removed.

nicklasl (approved):

vahidlazio (approved):

PR #124, opened 2024-03-01
feat!: Total Confidence - Eventsender

vahidlazio:


nicklasl (commented):
Great start! Left some comments and thoughts for discussion purposes.

Provider/src/main/java/com/spotify/confidence/ConfidenceExtensions.kt
```diff
@@ -0,0 +1,45 @@
...
+            OpenFeatureAPI.getEvaluationContext()?.let { evalContext ->
+                val map = mutableMapOf<String, EventValue>()
+                map["targeting_key"] = EventValue.String(evalContext.getTargetingKey())
+                evalContext.asMap().forEach {
+                    map[it.key] = EventValue.String(Json.encodeToString(it.value))
+                }
+                map
+            } ?: mapOf()
+        }
+    ),
```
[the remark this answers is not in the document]
vahidlazio: please consider that this is a lambda here and the function will be used in the event sender internal engine to be invoked and new list will be generated based on the current context. type `fields` is a lambda.

Provider/src/main/java/com/spotify/confidence/ConfidenceValue.kt
```diff
@@ -0,0 +1,18 @@
...
+@Serializable(ConfidenceValueSerializer::class)
+sealed class ConfidenceValue {
+    data class String(val string: kotlin.String) : ConfidenceValue()
+    data class Double(val double: kotlin.Double) : ConfidenceValue()
+    data class Boolean(val boolean: kotlin.Boolean) : ConfidenceValue()
+    data class Integer(val integer: Int) : ConfidenceValue()
+    data class Struct(val map: Map<kotlin.String, ConfidenceValue>) : ConfidenceValue()
+    data class List(val list: kotlin.collections.List<ConfidenceValue>) : ConfidenceValue()
+    data class Date(@Serializable(DateSerializer::class) val date: java.util.Date) : ConfidenceValue()
+
```
nicklasl, line 16: Missing `Timestamp`?

Provider/src/main/java/com/spotify/confidence/ConfidenceValue.kt
```diff
@@ -0,0 +1,18 @@
...
+@Serializable(ConfidenceValueSerializer::class)
+sealed class ConfidenceValue {
+    data class String(val string: kotlin.String) : ConfidenceValue()
+    data class Double(val double: kotlin.Double) : ConfidenceValue()
+    data class Boolean(val boolean: kotlin.Boolean) : ConfidenceValue()
+    data class Integer(val integer: Int) : ConfidenceValue()
+    data class Struct(val map: Map<kotlin.String, ConfidenceValue>) : ConfidenceValue()
+    data class List(val list: kotlin.collections.List<ConfidenceValue>) : ConfidenceValue()
+    data class Date(@Serializable(DateSerializer::class) val date: java.util.Date) : ConfidenceValue()
```
nicklasl, line 15: Check if this serializing is going to work with the backend.

Provider/src/main/java/com/spotify/confidence/CommonContextBuilder.kt
```diff
@@ -0,0 +1,65 @@
...
+import com.spotify.confidence.CommonContextKeys.DEVICE_MODEL_KEY
+import com.spotify.confidence.CommonContextKeys.DEVICE_NAME_KEY
+import com.spotify.confidence.CommonContextKeys.DEVICE_TYPE_KEY
+import com.spotify.confidence.CommonContextKeys.OS_NAME_KEY
+import com.spotify.confidence.CommonContextKeys.OS_VERSION_KEY
+import com.spotify.confidence.CommonContextKeys.SCREEN_DENSITY_KEY
+import com.spotify.confidence.CommonContextKeys.SCREEN_HEIGHT_KEY
+import com.spotify.confidence.CommonContextKeys.SCREEN_WIDTH_KEY
+
+internal fun Confidence.addCommonContext(context: Context): Confidence {
```
nicklasl, line 19: I suggest we defer this concept for later. Please pull out of the PR but maybe store it somewhere for later inspiration.

Provider/src/main/java/com/spotify/confidence/EventSender.kt
```diff
@@ -0,0 +1,40 @@
...
+import java.io.File
+
+interface EventSender : Contextual {
+    fun send(
+        definition: String,
+        payload: ConfidenceFieldsType = mapOf()
+    )
+
+    fun onLowMemory(body: (List<File>) -> Unit): Contextual
+    fun stop()
```
nicklasl, line 16: I think it would be best to keep this functionality on the root (parent) Confidence object. Otherwise someone could run this on a child event sender and make all the event sender break, right?

Provider/src/main/java/com/spotify/confidence/Confidence.kt
```diff
@@ -0,0 +1,157 @@
...
+                if (addCommonContext) {
+                    it.addCommonContext(context)
+                }
+            }
+        }
+    }
+}
+
+internal fun Map<String, ConfidenceValue>.openFeatureFlatten(): Map<String, ConfidenceValue> {
+    val context = toMutableMap()
```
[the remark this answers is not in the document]
nicklasl: To make it more readable.

Provider/src/main/java/com/spotify/confidence/ConfidenceContext.kt
```diff
@@ -0,0 +1,63 @@
...
+interface Contextual : ConfidenceContextProvider {
+    fun withContext(context: Map<String, ConfidenceValue>): Contextual
+
+    fun putContext(context: Map<String, ConfidenceValue>)
+    fun setContext(context: Map<String, ConfidenceValue>)
+    fun putContext(key: String, value: ConfidenceValue)
+    fun removeContext(key: String)
+}
+
+class CommonContext : ConfidenceContextProvider {
```
nicklasl, line 24: Is this unused?

Provider/src/main/java/com/spotify/confidence/ConfidenceContext.kt
```diff
@@ -0,0 +1,63 @@
...
+
+fun EventSender.putContext(context: Map<String, Any>) {
+    val serialized = JsonAnySerializer.encodeToString(context)
+    return putContext(Json.decodeFromString(serialized))
+}
+
+fun EventSender.putContext(key: String, value: Any) {
+    val serialized = JsonAnySerializer.encodeToString(value)
+    return putContext(key, Json.decodeFromString(serialized))
+}
```
nicklasl, line 63: I suggest we defer this concept for later. Please pull out of the PR but maybe store it somewhere for later inspiration.

Provider/src/main/java/com/spotify/confidence/ConfidenceFlagEvaluation.kt
```diff
@@ -0,0 +1,134 @@
...
+            value = defaultValue,
+            reason = ResolveReason.ERROR,
+            errorCode = ErrorCode.PROVIDER_NOT_READY
+        )
+    }
+    requireNotNull(this)
+    val resolvedFlag = this.flags.firstOrNull { it.flag == parsedKey.flagName }
+        ?: throw FlagNotFoundError(
+            "Could not find flag named: ${parsedKey.flagName}"
+        )
```
nicklasl, line 23: During testing we saw that this exception was thrown out of the Provider and made the app crash.
Most likely we need to wrap the `getEvaluation()` function when called in provider in a try catch and re-map the exception to OF exceptions for it to work correctly?

Provider/src/main/java/com/spotify/confidence/ConfidenceValue.kt
fabriziodemaria, line 15: Should we handle `Date` and `Timestamp` independently, since they are validated differently in the backend?

Provider/src/main/java/com/spotify/confidence/EventSender.kt
```diff
@@ -0,0 +1,40 @@
...
+        ?: JsonAnySerializer.encodeToString(payload)
+    send(definition, Json.decodeFromString(serializedT))
+}
+
+inline fun <reified T> EventSender.send(
+    definition: String,
+    payload: T
+) {
+    send(definition, payload, null)
+}
```
nicklasl, line 40: I suggest we defer this concept for later. Please pull out of the PR but maybe store it somewhere for later inspiration.

Provider/src/main/java/com/spotify/confidence/CommonContextKeys.kt
```diff
@@ -0,0 +1,38 @@
+package com.spotify.confidence
+object CommonContextKeys {
```
fabriziodemaria, line 2: Would be nice to have this in a separate PR, and perhaps be worked on in conjunction with iOS to make sure all of these fields work well in "all" mobile paltforms

Provider/src/main/java/com/spotify/confidence/EventSenderEngine.kt
```diff
@@ -0,0 +1,121 @@
...
+                        }
+                    }
+                }
+            }
+        }
+    }
+
+    override fun onLowMemoryChannel(): Channel<List<File>> {
+        return eventStorage.onLowMemoryChannel()
+    }
```
nicklasl, line 83: I suggest we defer this concept for later. Please pull out of the PR but maybe store it somewhere for later inspiration.

Provider/src/main/java/com/spotify/confidence/EventSenderUploader.kt
```diff
@@ -0,0 +1,74 @@
...
+import okhttp3.Request
+import okhttp3.RequestBody.Companion.toRequestBody
+import java.util.Date
+
+internal interface EventSenderUploader {
+    suspend fun upload(events: EventBatch): Boolean
+}
+
+@Serializable
+internal data class EventBatch(
```
nicklasl, line 22: We should check with the backend but I think we should also be able to send the `sdk` data here.
From java:
```
Sdk.newBuilder()
            .setId(SdkId.SDK_ID_JAVA_CONFIDENCE)
            .setVersion(SdkUtils.getSdkVersion())
            .build();
```

Provider/src/main/java/com/spotify/confidence/RemoteFlagResolver.kt
```diff
@@ -0,0 +1,96 @@
...
+    val evaluationContext: Map<
+        String,
+        @Serializable(ConfidenceValueSerializer::class) ConfidenceValue>,
+    val clientSecret: String,
+    val apply: Boolean,
+    val sdk: Sdk
+)
+
+private fun Response.toResolveFlags(): ResolveResponse {
+    val bodyString = body!!.string()
```
nicklasl, line 87: `!!` 😱

Provider/src/main/java/com/spotify/confidence/client/Types.kt
```diff
@@ -55,4 +47,14 @@ sealed interface SchemaType {
     object DoubleSchema : SchemaType
     object StringSchema : SchemaType
     object BoolSchema : SchemaType
-}
\ No newline at end of file
+}
+
+sealed class ResolveResponse {
+    object NotModified : ResolveResponse()
```
nicklasl, line 53: Not used any more?

Provider/src/main/java/com/spotify/confidence/client/network/ApplyFlagsInteractor.kt
```diff
@@ -52,17 +47,9 @@ internal class ApplyFlagsInteractorImpl(
...
     val clientSecret: String,
     val resolveToken: String,
     val sdk: Sdk
-)
-
-private val json = Json {
-    serializersModule = SerializersModule {
-        contextual(UUIDSerializer)
-        contextual(DateSerializer)
-        contextual(StructureSerializer)
```
nicklasl, line 66: Do we not need these serializer?

Provider/src/main/java/com/spotify/confidence/client/serializers/Serializers.kt
```diff
@@ -25,79 +23,103 @@ import kotlinx.serialization.json.Json
...
+        when (val element = decoder.decodeJsonElement()) {
+            is JsonPrimitive -> {
+                val jsonPrimitive = element.jsonPrimitive
+                return when {
+                    jsonPrimitive.isString -> ConfidenceValue.String(jsonPrimitive.content)
+                    jsonPrimitive.booleanOrNull != null -> ConfidenceValue.Boolean(jsonPrimitive.boolean)
+                    jsonPrimitive.intOrNull != null -> ConfidenceValue.Integer(jsonPrimitive.int)
+                    jsonPrimitive.doubleOrNull != null -> ConfidenceValue.Double(jsonPrimitive.double)
+                    else -> ConfidenceValue.Null
+                }
```
nicklasl, line 78: as discussed, these are pretty uncertain, let's defer until we've aligned on how to handle.

Provider/src/main/java/com/spotify/confidence/client/serializers/Serializers.kt
```diff
@@ -150,25 +172,25 @@ internal object NetworkResolvedFlagSerializer : KSerializer<ResolvedFlag> {
...
                 Json.decodeFromString(SchemaTypeSerializer, schemasJson.toString())
             val valueJson = json["value"].toString()
-            val values: Structure =
+            val values: ConfidenceValue.Struct =
                 Json.decodeFromString(FlagValueSerializer(flagSchema), valueJson)
 
-            if (flagSchema.schema.size != values.asMap().size) {
-                throw OpenFeatureError.ParseError("Unexpected flag name in resolve flag data: $flag")
+            if (flagSchema.schema.size != values.map.size) {
+                throw ParseError("Unexpected flag name in resolve flag data: $flag")
```
nicklasl, line 179: Here's another of the exceptions that will escape OpenFeatureApi.

Provider/src/main/java/com/spotify/confidence/ConfidenceValue.kt
nicklasl: most likely it requires `YYYY-MM-DD`

Provider/src/test/java/com/spotify/confidence/EventSenderIntegrationTest.kt
```diff
@@ -0,0 +1,191 @@
...
+            Assert.assertEquals(totalFiles.iterator().asSequence().toList().size, 1)
+
+            val currentFile = directory
+                .walkFiles()
+                .filter { !it.name.endsWith("ready") }
+                .first()
+            Assert.assertEquals(eventStorage.eventsFor(currentFile).size, 2)
+        }
+    }
+}
```
nicklasl, line 191: I'm missing tests that validates that parent-child context inheritance + removal etc works. Or did I miss them?

Provider/src/main/java/com/spotify/confidence/ConfidenceFlagEvaluation.kt
```diff
@@ -0,0 +1,134 @@
...
+
+    // TODO revisit where to apply flags
```
fabriziodemaria, line 25: Shall we address this TODO? From a quick glance it doesn't seem like the apply logic has been changed

Provider/src/main/java/com/spotify/confidence/Confidence.kt
```diff
@@ -0,0 +1,157 @@
+package com.spotify.confidence
```
fabriziodemaria, line 1: Shall we consider splitting the Confidence stuff into a separate module early? Would doing this later cause friction for our users, when updating our library?

Provider/src/main/java/com/spotify/confidence/ConfidenceValue.kt
vahidlazio: created a [ticket here](https://spotify.atlassian.net/browse/HAW-519)

Provider/src/main/java/com/spotify/confidence/Confidence.kt
```diff
@@ -0,0 +1,157 @@
...
+        eventSenderEngine.stop()
+    }
+
+    companion object {
+        fun create(
+            context: Context,
+            clientSecret: String,
+            region: ConfidenceRegion = ConfidenceRegion.GLOBAL,
+            dispatcher: CoroutineDispatcher = Dispatchers.IO,
+            addCommonContext: Boolean = true
```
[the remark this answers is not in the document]
vahidlazio: will be moved to a new PR.

Provider/src/main/java/com/spotify/confidence/ConfidenceValue.kt
vahidlazio: created a ticket to handle this. [here](https://spotify.atlassian.net/browse/HAW-519)

Provider/src/main/java/com/spotify/confidence/ConfidenceValue.kt
vahidlazio: we used the same serializer as we have in main for type `Value` . i recommend to tackle this if needed in the ticket mentioned above.

Provider/src/main/java/com/spotify/confidence/CommonContextBuilder.kt
vahidlazio: will be moved.

Provider/src/main/java/com/spotify/confidence/ConfidenceContext.kt
vahidlazio: will be moved.

Provider/src/main/java/com/spotify/confidence/EventSender.kt
vahidlazio: I created a `RootConfidence` and `ChildConfidence` and moved the stop functionality to the RootConfidence only like we did in Java. but I don't like this approach so much, I'd rather have each instance signal its own shutdown and the main engine stops what it's doing if all the instances are shut down instead. which I'll tackle it in another PR. is having this `RootConfidence` which its only difference with `Confidence` interface is one additional `shutdown` function good enough for this PR?

Provider/src/main/java/com/spotify/confidence/EventStorage.kt
```diff
@@ -0,0 +1,125 @@
...
+    private fun getFileWithName(name: String): File {
+        val directory = context.getDir(DIRECTORY, Context.MODE_PRIVATE)
+        return File(directory, name)
+    }
+
+    companion object {
+        const val DIRECTORY = "events"
+        const val READY_TO_SENT_EXTENSION = ".ready"
+        const val EVENT_WRITE_DELIMITER = ",\n"
+        const val MAX_STORAGE: Long = 4L * 1024 * 1024 // 4MB
```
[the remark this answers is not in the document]
vahidlazio: onMemory concept will be tackled in a separate PR.

Provider/src/main/java/com/spotify/confidence/client/network/ApplyFlagsInteractor.kt
vahidlazio: now they are handled in the data class declaration instead of declaring them in the Json object like here.

Provider/src/main/java/com/spotify/confidence/client/serializers/Serializers.kt
vahidlazio: updated to use the `Polymorphic` serialization like we do for `Value` in the sdk.

Provider/src/main/java/com/spotify/confidence/ConfidenceFlagEvaluation.kt
vahidlazio: todo is removed.

Provider/src/test/java/com/spotify/confidence/EventSenderIntegrationTest.kt
vahidlazio: done.

Provider/src/main/java/com/spotify/confidence/Confidence.kt
```diff
@@ -0,0 +1,185 @@
...
+
+internal const val OPEN_FEATURE_CONTEXT_KEY = "open_feature"
+
+object ConfidenceFactory {
+    fun create(
+        context: Context,
+        clientSecret: String,
+        region: ConfidenceRegion = ConfidenceRegion.GLOBAL,
+        dispatcher: CoroutineDispatcher = Dispatchers.IO
+    ): RootConfidence {
```
nicklasl, line 153: What do you think about returning `Confidence` here and have `Confidence` also include the `shutdown()`.
However we keep the two different types (`RootConfidence` and `ChildConfidence`) but running `shutdown()` on a child is a no-op (only log).

Provider/src/main/java/com/spotify/confidence/ConfidenceFeatureProvider.kt
```diff
@@ -140,102 +127,66 @@ class ConfidenceFeatureProvider private constructor(
...
-                )
-            }
+        try {
+            return providerCache.get().getEvaluation(
+                key,
+                defaultValue,
+                confidence.getContext().openFeatureFlatten()
+            ) { flagName, resolveToken ->
+                // this lambda will be invoked inside the evaluation process
+                // and only if the resolve reason is not targeting key error.
```
nicklasl, line 183: 👏

Provider/src/main/java/com/spotify/confidence/ConfidenceValue.kt
```diff
@@ -0,0 +1,26 @@
...
+
+    @Serializable data class Boolean(val boolean: kotlin.Boolean) : ConfidenceValue
+
+    @Serializable data class Integer(val integer: Int) : ConfidenceValue
+
+    @Serializable data class Struct(val map: Map<kotlin.String, ConfidenceValue>) : ConfidenceValue
+
+    @Serializable data class List(val list: kotlin.collections.List<ConfidenceValue>) : ConfidenceValue
+
+    @Serializable data class Date(@Serializable(DateSerializer::class) val date: java.util.Date) : ConfidenceValue
```
nicklasl, line 23: I'm fairly certain that this format (defined by this serializer) is not what is expected for a Date type in an event schema. Did you try this against the backend?

Provider/src/main/java/com/spotify/confidence/ConfidenceValue.kt
```diff
@@ -0,0 +1,26 @@
+package com.spotify.confidence
+
+import com.spotify.confidence.client.serializers.ConfidenceValueSerializer
+import dev.openfeature.sdk.DateSerializer
+import kotlinx.serialization.Serializable
+
+@Serializable(ConfidenceValueSerializer::class)
+sealed interface ConfidenceValue {
```
nicklasl, line 8: nit: We're missing the Timestamp type.

Provider/src/main/java/com/spotify/confidence/EventSender.kt
```diff
@@ -0,0 +1,16 @@
...
+interface EventSender : Contextual {
+    fun send(
+        definition: String,
+        payload: ConfidenceFieldsType = mapOf()
+    )
+
+    override fun withContext(context: Map<String, ConfidenceValue>): EventSender
+}
+
+interface FlushPolicy {
```
nicklasl, line 12: Can be `internal`.

Provider/src/main/java/com/spotify/confidence/NetworkConfidenceValueSerializer.kt
```diff
@@ -0,0 +1,69 @@
...
+                ConfidenceValue.Struct(value.map)
+            )
+
+            is ConfidenceValue.List -> encoder.encodeSerializableValue(
+                ListSerializer(NetworkConfidenceValueSerializer),
+                value.list
+            )
+
+            is ConfidenceValue.Date -> encoder.encodeSerializableValue(
+                DateSerializer,
```
nicklasl, line 64: Probably needs a Date-serializer that only uses YYYY-mm-dd?

Provider/src/main/java/com/spotify/confidence/ConfidenceFeatureProvider.kt
```diff
@@ -244,89 +195,92 @@ class ConfidenceFeatureProvider private constructor(
...
         ): Boolean {
-            val storage = StorageFileCache.create(context)
+            val storage = FileDiskStorage.create(context)
             val data = storage.read()
-            return data == null
+            return data == FlagResolution.EMPTY
         }
 
         @Suppress("LongParameterList")
         fun create(
```
nicklasl, line 204: Can we have a deprecated, backwards compatbile `create` fun that works as before with the `clientSecret` and `region`?

Provider/src/main/java/com/spotify/confidence/Confidence.kt
vahidlazio: then we can remove root and child confidence and have shutdown in confidence always. right now it could be no-op but in future it could de-register child and all children are de-registered then the engine will stop.

Provider/src/main/java/com/spotify/confidence/ConfidenceValue.kt
vahidlazio: is the Date type in event schema different than flag type for date? or evaluation context? because they should be the same, since we are using the same types. if they are different backend should change either for flags or events to support one single format.

Provider/src/test/java/com/spotify/confidence/DateSerializersTest.kt
```diff
@@ -0,0 +1,36 @@
...
+
+    @Test
+    fun date_time_serializer_works_with_timestamp_format() {
+        val pattern = "^\\d{4}-\\d{2}-\\d{2}T\\d{2}:\\d{2}:\\d{2}(\\.\\d{3})?(Z|[+-]\\d{2}:\\d{2})?\$"
+        val regex = Pattern.compile(pattern)
+        val now = Date()
+        val serializedDate = Json.encodeToString(DateTimeSerializer, now)
+        Assert.assertTrue(regex.matcher(serializedDate.removeSurrounding("\"")).matches())
+        val deserializedDate = Json.decodeFromString(DateTimeSerializer, serializedDate)
+        Assert.assertTrue(deserializedDate == now)
```
nicklasl, line 34: Can we make these "hard coded" instead?
Creating a date object for a _specific_ date and timezone and then verifying that this is what the `deserializedDate` is?

nicklasl (approved):

nicklasl (approved):

PR #129, opened 2024-04-17
fix: ConfidenceValue.List can only contain single type

vahidlazio:


Provider/src/main/java/com/spotify/confidence/ConfidenceFeatureProvider.kt
```diff
@@ -233,7 +233,14 @@ internal fun Value.toConfidenceValue(): ConfidenceValue = when (this) {
     is Value.Date -> ConfidenceValue.Timestamp(this.date)
     is Value.Double -> ConfidenceValue.Double(this.double)
     is Value.Integer -> ConfidenceValue.Integer(this.integer)
-    is Value.List -> ConfidenceValue.List(this.list.map { it.toConfidenceValue() })
+    is Value.List -> {
+        // if types are different, return an empty list
+        if (this.list.map { it.javaClass.simpleName }.groupBy { it }.size > 1) {
+            ConfidenceValue.List(listOf())
```
nicklasl, line 239: I really wish we could log something here.

nicklasl (approved):

PR #130, opened 2024-04-17
feat: include persistent visitor id in context

nicklasl:
While doing this I met with the issue of a stale context in a test.
Me and @vahidlazio decided that it made most sense with regards to our discussions this week to send back the actual value but also to signal in the `errorCode` that the evaluation may be stale.
The Open feature provider however "swallows" that error code.

### DIFF
--- a/Provider/src/main/java/com/spotify/confidence/VisitorUtil.kt
+++ b/Provider/src/main/java/com/spotify/confidence/VisitorUtil.kt
@@ -1,0 +1,22 @@
+package com.spotify.confidence
+
+import android.content.Context
+import java.util.UUID
+
+internal const val SHARED_PREFS_NAME = "confidence-visitor"
+internal const val VISITOR_ID_SHARED_PREFS_KEY = "visitorId"
+internal const val DEFAULT_VALUE = "unable-to-read"
+
+internal object VisitorUtil {
+    fun getId(context: Context): String {
+        return with(context.getSharedPreferences(SHARED_PREFS_NAME, Context.MODE_PRIVATE)) {
+            if (contains(VISITOR_ID_SHARED_PREFS_KEY)) {
+                getString(VISITOR_ID_SHARED_PREFS_KEY, DEFAULT_VALUE) ?: DEFAULT_VALUE
+            } else {
+                val visitorId = UUID.randomUUID().toString()
+                edit().putString(VISITOR_ID_SHARED_PREFS_KEY, visitorId).apply()
+                visitorId
+            }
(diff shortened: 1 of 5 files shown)


Provider/src/main/java/com/spotify/confidence/VisitorUtil.kt
```diff
@@ -0,0 +1,22 @@
...
+import android.content.Context
+import java.util.UUID
+
+internal const val SHARED_PREFS_NAME = "confidence-visitor"
+internal const val VISITOR_ID_SHARED_PREFS_KEY = "visitorId"
+internal const val DEFAULT_VALUE = "unable-to-read"
+
+internal object VisitorUtil {
+    fun getId(context: Context): String {
+        return with(context.getSharedPreferences(SHARED_PREFS_NAME, Context.MODE_PRIVATE)) {
```
fabriziodemaria, line 12: Let's make sure this stuff is not synced via the Cloud

vahidlazio: i think the way we fixed that in rcs was the backup tag in the manifest file, we can have that in our manifest file, but it’s always overridable from the host app because manifests will be merged and the host app takes priority. so it all depends on the host app. we can add this in our docs to warn the users about this.

fabriziodemaria: I think we should be in control of this behaviour

vahidlazio: this behaviour is always overridable through the host app as in android manifests are merged. even when we create rules for backup, this rules can be overridden in the host app.

vahidlazio: like [here ](https://developer.android.com/guide/topics/data/autobackup#:~:text=By%20default%2C%20Auto%20Backup%20includes,or%20getDir(String%2C%20int)) we can create custom rules to exclude the sharedPref and directories from backup and then tell the users about it.

fabriziodemaria: We wouldn't have this problem if we used a file, right?

vahidlazio: we do. all the files are backed up including the shared pref if they are not customised and if the backup full content is true in the manifest.

fabriziodemaria (approved):

PR #131, opened 2024-04-18
feat: Track app lifecycle, track activities lifecycle, track deeplink, track install/update

vahidlazio:


Provider/build.gradle.kts
```diff
@@ -55,6 +55,7 @@ dependencies {
     implementation(
         "org.jetbrains.kotlinx:kotlinx-serialization-json:${Versions.kotlinxSerialization}"
     )
+    implementation("androidx.lifecycle:lifecycle-process:2.6.1")
```
nicklasl, line 58: Thinking out loud:
Now this may be a library that already would exist in the host app, but I feel we need some way to measure binary increase for this type of changes.
Ideally we would ship "integrations/add-ons" like this in a separate artifact, like `confidence-analytics-lifecycle`.

Provider/src/main/java/com/spotify/confidence/AndroidLifecycleEventProducer.kt
```diff
@@ -0,0 +1,235 @@
...
+    override fun contextChanges(): Flow<Map<String, ConfidenceValue>> = contextFlow
+    override fun stop() {
+        application.unregisterActivityLifecycleCallbacks(this)
+        coroutineScope.launch(Dispatchers.Main) {
+            lifecycle.removeObserver(this@AndroidLifecycleEventProducer)
+        }
+    }
+
+    companion object {
+        private const val IS_FOREGROUND_KEY = "is_foreground"
```
nicklasl, line 178: Note: This is a good example of a context that we probably don't want to re-resolve flags for, right?

Provider/src/main/java/com/spotify/confidence/AndroidLifecycleEventProducer.kt
```diff
@@ -0,0 +1,235 @@
...
+        }
+
+        // setup lifecycle listeners
+        application.registerActivityLifecycleCallbacks(this)
+        lifecycle = ProcessLifecycleOwner.get().lifecycle
+        coroutineScope.launch(Dispatchers.Main) {
+            lifecycle.addObserver(this@AndroidLifecycleEventProducer)
+        }
+    }
+    override fun onActivityCreated(p0: Activity, p1: Bundle?) {
```
nicklasl, line 52: Some funky parameter namings througout this class.

Provider/src/main/java/com/spotify/confidence/AndroidLifecycleEventProducer.kt
```diff
@@ -0,0 +1,235 @@
...
+    private val lifecycle: Lifecycle
+
+    private val coroutineScope = CoroutineScope(Dispatchers.IO)
+
+    init {
+        val packageManager: PackageManager = application.packageManager
+        packageInfo = try {
+            packageManager.getPackageInfo(application.packageName, 0)
+        } catch (e: PackageManager.NameNotFoundException) {
+            throw Error("can't read package")
```
nicklasl, line 42: IMO we shouldn't throw here but just make `packageInfo` nullable.

Provider/src/main/java/com/spotify/confidence/Confidence.kt
```diff
@@ -88,6 +86,38 @@ class Confidence internal constructor(
...
+
+    override fun stop() {
+        for (producer in eventProducers) {
+            producer.stop()
+        }
+        if (parent == null) {
+            eventSenderEngine.stop()
+        } else {
+            // no-op for child confidence
+            coroutineScope.cancel()
```
nicklasl, line 118: shouldn't this also happen for `parent==null`?

Provider/src/main/java/com/spotify/confidence/EventSender.kt
```diff
@@ -6,11 +6,15 @@ interface EventSender : Contextual {
         payload: ConfidenceFieldsType = mapOf()
     )
 
+    fun track(eventProducer: EventProducer)
```
nicklasl, line 9: Food for thought: should it be possible to untrack a producer?

Provider/src/main/java/com/spotify/confidence/AndroidLifecycleEventProducer.kt
```diff
@@ -0,0 +1,235 @@
...
+    }
+
+    companion object {
+        private const val IS_FOREGROUND_KEY = "is_foreground"
+        private const val APP_VERSION = "AppVersion"
+        private const val APP_BUILD = "APP_BUILD"
+        private const val LEGACY_APP_BUILD = "LEGACY_APP_BUILD"
+        private const val APP_INSTALLED_EVENT = "app-installed"
+        private const val APP_UPDATED_EVENT = "app-updated"
+        private const val APP_CREATED = "app-on-created"
```
nicklasl, line 184:
```suggestion
        private const val APP_LAUNCHED = "app-launched"
```

Provider/build.gradle.kts
vahidlazio: yeah for sure we should separate the producers.

Provider/src/main/java/com/spotify/confidence/AndroidLifecycleEventProducer.kt
vahidlazio: yes but after merging your changes we shouldn't be returning stale anymore and i think it's still nice to send these.

Provider/src/main/java/com/spotify/confidence/Confidence.kt
vahidlazio: yes, you are right.

Provider/src/main/java/com/spotify/confidence/EventSender.kt
vahidlazio: i thought about it but then it should probably be named or return an id or a handler to untrack etc. was overkill for a draft.

Provider/build.gradle.kts
vahidlazio: created a ticket [here](https://spotify.atlassian.net/browse/HAW-539)

Provider/src/main/java/com/spotify/confidence/AndroidLifecycleEventProducer.kt
```diff
@@ -0,0 +1,235 @@
...
+            lifecycle.addObserver(this@AndroidLifecycleEventProducer)
+        }
+    }
+    override fun onActivityCreated(activity: Activity, bundle: Bundle?) {
+        trackDeepLink(activity)
+        activity.trackActivity("activity-created")
+    }
+
+    override fun onCreate(owner: LifecycleOwner) {
+        trackApplicationLifecycleEvents()
```
fabriziodemaria, line 58: Would it make sense to add a `APP_STARTED` event here? The challenge is to not include this into the context though

fabriziodemaria: Never mind, I see we have `APP_LAUNCHED`

Provider/src/main/java/com/spotify/confidence/AndroidLifecycleEventProducer.kt
```diff
@@ -0,0 +1,237 @@
...
+        private const val APP_VERSION = "APP_VERSION"
+        private const val APP_BUILD = "APP_BUILD"
+        private const val LEGACY_APP_BUILD = "LEGACY_APP_BUILD"
+
+        // Context keys
+        private const val IS_FOREGROUND_KEY = "is_foreground"
+
+        // Event keys
+        private const val APP_INSTALLED_EVENT = "app-installed"
+        private const val APP_UPDATED_EVENT = "app-updated"
```
fabriziodemaria, line 185: While manual testing, I noticed these are fired more than once in a single session

fabriziodemaria:
Removed activity events (a bit too verbose, we can think how to add them later if needed).

Added managed data:
        // Context keys
        IS_FOREGROUND_KEY = "is_foreground"

        // Event keys
        APP_INSTALLED_EVENT = "app-installed"
        APP_UPDATED_EVENT = "app-updated"
        APP_LAUNCHED_EVENT = "app-launched"

fabriziodemaria (approved):

fabriziodemaria (approved):

PR #133, opened 2024-04-22
feat: listen for changes in the provider for the context changes

vahidlazio:


Provider/src/main/java/com/spotify/confidence/ConfidenceFeatureProvider.kt
```diff
@@ -49,52 +51,58 @@ class ConfidenceFeatureProvider private constructor(
         }
     }
 
+    private fun startListeningForContext() {
+        coroutineScope.launch {
+            confidence.contextChanges
+                .drop(1)
+                .distinctUntilChanged()
+                .collect {
+                    resolve(InitialisationStrategy.FetchAndActivate)
```
nicklasl, line 60: What if the strategy was set to `ActivateAndFetch`? Somewhat makes me question the whole thing about fetch strategy.

vahidlazio: [here](https://github.com/spotify/confidence-openfeature-provider-kotlin/blob/cf8f712a141862f8d54f1c70593625b65770d754/Provider/src/main/java/com/spotify/confidence/ConfidenceFeatureProvider.kt#L71) we are resolving first with the current context and the given initialization strategy. then we start listening for the changes, if any change is happening, we refresh the cache after resolving (same logic as in main). I moved the drop and distinct to the confidence object so it's more clear that the `contextChanges` is only about changes from this point of time onwards.

Provider/src/main/java/com/spotify/confidence/ConfidenceFeatureProvider.kt
```diff
@@ -49,52 +49,59 @@ class ConfidenceFeatureProvider private constructor(
...
-        strategy: InitialisationStrategy
-    ) {
-        // refresh cache with the last stored data
-        storage.read().let(providerCache::refresh)
-        if (strategy == InitialisationStrategy.ActivateAndFetchAsync) {
-            eventHandler.publish(OpenFeatureEvents.ProviderReady)
-        }
+    private suspend fun resolve(strategy: InitialisationStrategy) {
+        try {
+            val resolveResponse = confidence.resolve(listOf())
```
nicklasl, line 82: I assume we don't have `RECONCILING` support from the SDK yet?
Otherwise we would emit that before resolve, right?

Provider/src/main/java/com/spotify/confidence/ConfidenceFeatureProvider.kt
```diff
@@ -106,13 +113,15 @@ class ConfidenceFeatureProvider private constructor(
...
-        if (newContext != oldContext) {
-            // on the new context we want to fetch new values and update
-            // the storage & cache right away which is why we pass `InitialisationStrategy.FetchAndActivate`
-            internalInitialize(
-                newContext,
-                InitialisationStrategy.FetchAndActivate
-            )
+        val context = newContext.toConfidenceContext()
+        for (entry in context.map) {
+            confidence.putContext(entry.key, entry.value)
```
nicklasl, line 118: calling this in a for loop will emit multiple times on the `contextChanges` Flow. I don't think we want that.
Better to collect all of these changes in one emit.

Provider/src/test/java/com/spotify/confidence/ConfidenceFeatureProviderTests.kt
```diff
@@ -587,6 +587,7 @@ internal class ConfidenceFeatureProviderTests {
         )
```
nicklasl, line 587: We need tests for the `contextChanges` behaviour.

vahidlazio: [done](https://github.com/spotify/confidence-openfeature-provider-kotlin/pull/133/commits/b445658db9247d6a7452c167d41ecd446622a8d4)

Provider/src/main/java/com/spotify/confidence/ConfidenceFeatureProvider.kt
vahidlazio: done in the last commit. made the update with removed keys transactional.

Provider/src/main/java/com/spotify/confidence/ConfidenceFeatureProvider.kt
```diff
@@ -106,14 +111,9 @@ class ConfidenceFeatureProvider private constructor(
...
-            // on the new context we want to fetch new values and update
-            // the storage & cache right away which is why we pass `InitialisationStrategy.FetchAndActivate`
-            internalInitialize(
-                newContext,
-                InitialisationStrategy.FetchAndActivate
-            )
-        }
+        val context = newContext.toConfidenceContext()
+        val removedKeys = oldContext?.asMap()?.keys?.minus(newContext.asMap().keys) ?: emptySet()
+        confidence.putContext(context.map, removedKeys.toList())
```
fabriziodemaria, line 116: Just a thought: could we updated `removedKeys` from here and just call `confidence.putContext(context.map)`? `newContext` should already not include any key that is in the `removedKeys` list

Provider/src/main/java/com/spotify/confidence/Confidence.kt
```diff
@@ -40,34 +50,50 @@ class Confidence internal constructor(
...
+        }
+        this.removedKeys.addAll(removedKeys)
+        contextMap.value = map
     }
 
+    @Synchronized
     override fun removeContext(key: String) {
+        val map = contextMap.value.toMutableMap()
+        map.remove(key)
         removedKeys.add(key)
```
fabriziodemaria, line 89: In Swift, the removed keys are rather stored in Confidence: https://github.com/spotify/confidence-openfeature-provider-swift/blob/0d1cefdeb766a3d24c7d05be5f834d8855f271f3/Sources/Confidence/Confidence.swift#L101C24-L101C42
Let's align on one approach

Provider/src/main/java/com/spotify/confidence/ConfidenceFeatureProvider.kt
fabriziodemaria: In Swift we don't force this to be `FetchAndActivate`: https://github.com/spotify/confidence-openfeature-provider-swift/blob/0d1cefdeb766a3d24c7d05be5f834d8855f271f3/Sources/ConfidenceProvider/ConfidenceFeatureProvider.swift#L107
Is it expected? 
(I am also noticing that the `strategy` argument is not used in the linked Swift function)

Provider/src/main/java/com/spotify/confidence/ConfidenceFeatureProvider.kt
```diff
@@ -49,52 +49,57 @@ class ConfidenceFeatureProvider private constructor(
...
-            )
+            // refresh cache with the last stored data
+            storage.read().let(providerCache::refresh)
+            if (initialisationStrategy == InitialisationStrategy.ActivateAndFetchAsync) {
+                eventHandler.publish(OpenFeatureEvents.ProviderReady)
+            }
+
+            coroutineScope.launch(networkExceptionHandler) {
+                val context = initialContext.toConfidenceContext()
+                confidence.putContext(context.map)
```
fabriziodemaria, line 71: In Swift we rather update the context in Confidence directly, and then read from it: https://github.com/spotify/confidence-openfeature-provider-swift/blob/0d1cefdeb766a3d24c7d05be5f834d8855f271f3/Sources/ConfidenceProvider/ConfidenceFeatureProvider.swift#L87

I don't think this is necessarily a blocker, but maintaining a more aligned implementation will make things easier for us

fabriziodemaria (commented):
Mostly minor comments bringing up comparisons with the Swift counterpart PR

Provider/src/main/java/com/spotify/confidence/ConfidenceFeatureProvider.kt
vahidlazio: that's a bug in swift, i fixed it [here](https://github.com/spotify/confidence-openfeature-provider-swift/pull/101), good catch!

Provider/src/main/java/com/spotify/confidence/ConfidenceFeatureProvider.kt
vahidlazio: I don't think we can: `putContext` just puts this keys in the confidence context map, if we had a, b, c and now new context has d, if we put d we are going to have a, b, c, d. consider that putContext doesn't replace the whole context.

Provider/src/main/java/com/spotify/confidence/Confidence.kt
vahidlazio: it's the same, this class is confidence. the link in swift the `confidence` variable is something that comes from the confidence. maybe the naming is a bit confusing over there.

Provider/src/main/java/com/spotify/confidence/ConfidenceFeatureProvider.kt
vahidlazio: > we rather update the context in Confidence directly, and then read from it:
we do the same here i suppose, could you explain more what's the difference?

fabriziodemaria: We do the same, you are right

fabriziodemaria (approved):

PR #134, opened 2024-04-23
refactor: Add `message` container to `payload`

fabriziodemaria:
Also minor changes in the various interfaces and how events are stored on disk

### DIFF
--- a/Provider/src/main/java/com/spotify/confidence/EventSenderUploader.kt
+++ b/Provider/src/main/java/com/spotify/confidence/EventSenderUploader.kt
@@ -52,6 +52,7 @@ internal class EventSenderUploaderImpl(
     }
 
     override suspend fun upload(events: EventBatchRequest): Boolean = withContext(dispatcher) {
+        Event
         val networkJson = Json {
             serializersModule = SerializersModule {
                 contextual(NetworkConfidenceValueSerializer)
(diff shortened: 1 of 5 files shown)


Provider/src/main/java/com/spotify/confidence/EventSenderUploader.kt
```diff
@@ -52,11 +52,13 @@ internal class EventSenderUploaderImpl(
     }
 
     override suspend fun upload(events: EventBatchRequest): Boolean = withContext(dispatcher) {
+        Event
```
nicklasl, line 55: 👀

fabriziodemaria: Wha.... What happened there?? And how is this still compiling :D

Provider/src/test/java/com/spotify/confidence/EventSenderIntegrationTest.kt
```diff
@@ -124,6 +124,68 @@ class EventSenderIntegrationTest {
...
+        )
+        engine.emit(
+            eventName = "my_event",
+            message = mapOf(
+                "a" to ConfidenceValue.Integer(0),
+                "message" to ConfidenceValue.Integer(1)
+            ),
+            context = mapOf(
+                "a" to ConfidenceValue.Integer(2),
+                "message" to ConfidenceValue.Integer(3)
```
nicklasl, line 169: I think this is going a bit overboard in this specific test.
Can we move it to a separate one titled something with `handles_message_key_collision`?

nicklasl (approved):

Provider/src/test/java/com/spotify/confidence/EventSenderIntegrationTest.kt
fabriziodemaria: I renamed the test

nicklasl (approved):

vahidlazio (approved):

PR #136, opened 2024-04-29
refactor!: split the confidence module to a new module

vahidlazio:


confidence/build.gradle.kts
```diff
@@ -0,0 +1,119 @@
...
+    testImplementation("org.jetbrains.kotlinx:kotlinx-coroutines-test:${Versions.coroutines}")
+    testImplementation("org.mockito.kotlin:mockito-kotlin:${Versions.kotlinMockito}")
+    testImplementation("com.squareup.okhttp3:mockwebserver:${Versions.mockWebServer}")
+}
+
+publishing {
+    publications {
+        register<MavenPublication>("release") {
+            groupId = project.extra["groupId"].toString()
+            artifactId = "spotify-confidence-kotlin"
```
nicklasl, line 65:
```suggestion
            artifactId = "spotify-confidence-android"
```

nicklasl:
When we merge this we need to include that it is a breaking change.

settings.gradle.kts
```diff
@@ -1,5 +1,7 @@
 import org.gradle.api.initialization.resolve.RepositoriesMode
 
+include(":confidence")
+
```
nicklasl, line 4: `rootProject.name` in this file needs update.

confidence/build.gradle.kts
```diff
@@ -0,0 +1,119 @@
...
+    id("com.android.library")
+    id("org.jetbrains.kotlin.android")
+    id("maven-publish")
+    id("signing")
+    kotlin("plugin.serialization").version("1.8.10").apply(true)
+}
+
+val providerVersion = project.extra["version"].toString()
+
+object Versions {
```
nicklasl, line 11: Ideally we want to keep these versions set in the project gradle file to share them

release-please-config.json
```diff
@@ -10,7 +10,8 @@
             "extra-files": [
```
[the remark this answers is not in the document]
nicklasl: We can take this later though

ConfidenceDemoApp/src/main/java/com/example/confidencedemoapp/MainVm.kt
```diff
@@ -30,51 +23,39 @@ class MainVm(app: Application) : AndroidViewModel(app) {
...
         mutableMap["hello"] = ConfidenceValue.Boolean(false)
         mutableMap["NN"] = ConfidenceValue.Double(20.0)
         mutableMap["list"] = ConfidenceValue.stringList(listOf(""))
         mutableMap["my_struct"] = ConfidenceValue.Struct(mapOf("x" to ConfidenceValue.Double(2.0)))
 
-        val confidence = ConfidenceFactory.create(
+        confidence = ConfidenceFactory.create(
             app.applicationContext,
             clientSecret,
+            initialContext = mapOf("targeting_key" to ConfidenceValue.String("a98a4291-53b0-49d9-bae8-73d3f5da2070")),
```
nicklasl, line 47: Would love for us to move away from `targeting_key`. Do we keep it for the flag rules in the backend expect it?

ConfidenceDemoApp/src/main/java/com/example/confidencedemoapp/MainVm.kt
```diff
@@ -30,51 +23,39 @@ class MainVm(app: Application) : AndroidViewModel(app) {
...
-                context = app.applicationContext,
-                initialisationStrategy = strategy
-            )
-            OpenFeatureAPI.setProviderAndWait(provider, Dispatchers.IO)
+            if(confidence.isStorageEmpty()) {
+                confidence.fetchAndActivate()
+            } else {
+                confidence.activate()
+                confidence.asyncFetch()
+            }
```
nicklasl, line 58: This signals to me that it would be nice to have an "entry function" for our users to perhaps pass a strategy?

confidence/src/main/java/com/spotify/confidence/Confidence.kt
```diff
@@ -114,12 +153,50 @@ class Confidence internal constructor(
...
+                    if (resolveResponse.data != FlagResolution.EMPTY) {
+                        diskStorage.store(resolveResponse.data)
+                    }
+                }
+            } catch (e: ParseError) {
+                throw ParseError(e.message)
+            }
+        }
+
+    fun activate() {
```
nicklasl, line 177: Is there a time when we do not want to activate?

ConfidenceDemoApp/src/main/java/com/example/confidencedemoapp/MainVm.kt
vahidlazio: I'm not sure. i think backend works without targeting key.

ConfidenceDemoApp/src/main/java/com/example/confidencedemoapp/MainVm.kt
vahidlazio: I actually think having control like this is simpler rather than understanding another concept like initializationStrategy, I believe most of the users will only use async fetch or activateAndFetch only.

confidence/src/main/java/com/spotify/confidence/Confidence.kt
vahidlazio: I think no but confidence object is dumb. I prefer these actions to be explicit and set by the user and be mentioned in the user guide.

confidence/build.gradle.kts
vahidlazio: exactly.

confidence/src/main/java/com/spotify/confidence/Confidence.kt
```diff
@@ -112,6 +146,43 @@ class Confidence internal constructor(
         eventSenderEngine.emit(eventName, message, getContext())
     }
 
+    private val networkExceptionHandler by lazy {
+        CoroutineExceptionHandler { _, _ ->
+            // network failed, provider is ready but with default/cache values
+            print("")
```
nicklasl, line 152: let's signal this in some better way?

confidence/src/main/java/com/spotify/confidence/RemoteFlagResolver.kt
```diff
@@ -19,7 +18,7 @@ import okhttp3.Request
 import okhttp3.RequestBody.Companion.toRequestBody
 import okhttp3.Response
 
-internal interface FlagResolver {
+interface FlagResolver {
```
nicklasl, line 21: why does it need to be public?

confidence/src/main/java/com/spotify/confidence/cache/FileDiskStorage.kt
```diff
@@ -10,7 +10,7 @@ import java.io.File
 internal const val FLAGS_FILE_NAME = "confidence_flags_cache.json"
 internal const val APPLY_FILE_NAME = "confidence_apply_cache.json"
 
-internal class FileDiskStorage private constructor(
+class FileDiskStorage private constructor(
```
nicklasl, line 13: preferably not public

confidence/build.gradle.kts
vahidlazio: yes.

ConfidenceDemoApp/src/main/java/com/example/confidencedemoapp/MainVm.kt
vahidlazio: yes, for sure, i would go only with fetchAndActivate.

confidence/src/main/java/com/spotify/confidence/Confidence.kt
vahidlazio: ah we don't need this, forgot to remove.

confidence/src/main/java/com/spotify/confidence/RemoteFlagResolver.kt
vahidlazio: we don't, reverted.

confidence/build.gradle.kts
```diff
@@ -0,0 +1,120 @@
...
+    publications {
+        register<MavenPublication>("release") {
+            groupId = project.extra["groupId"].toString()
+            artifactId = "spotify-confidence-android"
+            version = providerVersion
+
+            pom {
+                name.set("Confidence SDK Android")
+                description.set("Android SDK for Confidence")
+                url.set("https://github.com/spotify/confidence-openfeature-provider-kotlin")
```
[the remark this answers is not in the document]
vahidlazio: I open a new PR to address changes regarding the change of repo name.

Provider/src/test/java/com/spotify/confidence/openfeature/InMemorySharedPreferences.kt
```diff
@@ -0,0 +1,100 @@
+package com.spotify.confidence.openfeature
```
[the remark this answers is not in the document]
vahidlazio: we also have a integration test in OF which means we are also going to need the same fake.

confidence/src/main/java/com/spotify/confidence/Confidence.kt
```diff
@@ -19,41 +20,69 @@ import kotlinx.coroutines.flow.drop
 import kotlinx.coroutines.launch
 import okhttp3.OkHttpClient
 
+internal const val SDK_ID = "SDK_ID_KOTLIN_PROVIDER"
```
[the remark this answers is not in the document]
fabriziodemaria: We should use `SDK_ID_KOTLIN_CONFIDENCE` ([ref](https://ghe.spotify.net/konfidens/epx-flags-resolver/blob/10cef43bd3c931ae9d8f4e727ba13be3f71a31ea/epx-flags-resolver-lib/src/main/proto/confidence/flags/resolver/v1/types.proto#L70)). I agree also on keeping `SDK_ID_KOTLIN_PROVIDER` if it's used in the OF env

Provider/src/main/java/com/spotify/confidence/openfeature/ConfidenceFeatureProvider.kt
```diff
@@ -45,64 +41,24 @@ class ConfidenceFeatureProvider private constructor(
...
-                        providerCache.refresh(resolveResponse.data)
-                        eventHandler.publish(OpenFeatureEvents.ProviderReady)
-                    }
-
-                    InitialisationStrategy.ActivateAndFetchAsync -> {
-                        // do nothing
-                    }
+            InitialisationStrategy.FetchAndActivate -> {
+                coroutineScope.launch {
+                    confidence.fetchAndActivate()
```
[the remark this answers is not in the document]
vahidlazio: yes, it doesn't crash and it goes forward with activating what we have already in the storage.

ConfidenceDemoApp/build.gradle.kts
```diff
@@ -83,7 +83,7 @@ android {
 }
 
 dependencies {
-    implementation(project(":Provider"))
+    implementation(project(":confidence"))
```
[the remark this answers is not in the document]
vahidlazio: makes sense.

confidence/src/main/java/com/spotify/confidence/Confidence.kt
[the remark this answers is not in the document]
vahidlazio: changed to `SDK_ID_KOTLIN_CONFIDENCE`, however this sdkmetadata is internal to confidence, users from public api shouldn't be able to set if from the outside. we can have a metadata of the environment in which confidence is used to send **alongside** with the id and version to the backend.

ConfidenceDemoApp/src/main/java/com/example/confidencedemoapp/MainVm.kt
```diff
@@ -100,16 +85,14 @@ class MainVm(app: Application) : AndroidViewModel(app) {
...
-            )
+        val ctx = mapOf(
+            "user_id" to ConfidenceValue.String(UUID.randomUUID().toString()),
+            "picture" to ConfidenceValue.String("hej"),
+            "region" to ConfidenceValue.String("eu")
         )
         viewModelScope.launch {
             Log.d(TAG, "set new EvaluationContext")
-            OpenFeatureAPI.setEvaluationContext(ctx)
+            confidence.putContext(ctx)
```
fabriziodemaria, line 95: Do we have any way to know when the new values are available for the app to consume? 
In my testing the `refreshUI` that follows this line is always returning the older STALE value

fabriziodemaria (dismissed):

vahidlazio:
@fabriziodemaria I added `awaitReconciliation` function + an extension named `awaitPutContext`. I think that fixes the issue.

fabriziodemaria (approved):

PR #147, opened 2024-05-10
fix: do not throw while evaluating flag, return proper error

vahidlazio:


Confidence/src/main/java/com/spotify/confidence/ConfidenceFlagEvaluation.kt
```diff
@@ -19,9 +16,10 @@ fun <T> FlagResolution?.getEvaluation(
     }
     requireNotNull(this)
     val resolvedFlag = this.flags.firstOrNull { it.flag == parsedKey.flagName }
-        ?: throw FlagNotFoundError(
-            "Could not find flag named: ${parsedKey.flagName}",
-            parsedKey.flagName
+        ?: return Evaluation(
```
nicklasl, line 19: Do we need to check for this and throw in the provider instead now?

vahidlazio: probably, but then I was thinking what's the point of throwing of OF case? the sdk will just handle it and send default anyways. right?

nicklasl: I think the SDK expects the provider to throw for this case.

nicklasl: Never mind. The SDK can support that the provider throws but it does not require it.

nicklasl (approved):

PR #152, opened 2024-05-23
feat: add manual flush capabilities

nicklasl:
We want to allow for the user to manually trigger an upload to the backend services for all events emitted up to that point. 

The approach used in this PR is to treat a call to the `flush()` method as a special event emitted, which will mean that the events in the "write channel" will be guaranteed to have been written to disk (and thus applicable to be part of the "batch" to upload) before the actual flush happens.


fabriziodemaria (approved):
LGTM

PR #154, opened 2024-05-27
fix: Add version/build to context

fabriziodemaria:
`previous_version` and `previous_build` are not available in iOS: removing for consistency.
Also changing the context keys for version and build for consistency: https://github.com/spotify/confidence-sdk-swift/pull/127

### DIFF
--- a/Confidence/src/main/java/com/spotify/confidence/AndroidLifecycleEventProducer.kt
+++ b/Confidence/src/main/java/com/spotify/confidence/AndroidLifecycleEventProducer.kt
@@ -128,11 +128,6 @@ class AndroidLifecycleEventProducer(
         val currentVersion = ConfidenceValue.String(packageInfo?.versionName ?: "")
         val currentBuild = ConfidenceValue.String(packageInfo?.getVersionCode().toString() ?: "")
 
-        // Get the previous recorded version.
-        val previousVersion = sharedPreferences
-            .getString(APP_VERSION, null)
-            ?.let(ConfidenceValue::String)
-
         val previousBuild: ConfidenceValue.String? = sharedPreferences
             .getString(APP_BUILD, null)
             ?.let(ConfidenceValue::String)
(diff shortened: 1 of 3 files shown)


Confidence/src/main/java/com/spotify/confidence/AndroidLifecycleEventProducer.kt
```diff
@@ -141,14 +136,12 @@ class AndroidLifecycleEventProducer(
...
         } else if (currentBuild != previousBuild) {
             val message = mapOf(
-                "version" to currentVersion,
-                "build" to currentBuild,
-                "previous_version" to (previousVersion ?: ConfidenceValue.String("")),
-                "previous_build" to (previousBuild ?: ConfidenceValue.String(""))
+                APP_VERSION_KEY to currentVersion,
+                APP_BUILD_KEY to currentBuild
             )
             coroutineScope.launch { eventsFlow.emit(Event(APP_UPDATED_EVENT, message)) }
```
nicklasl, line 146: why do we sent these fields in the message part? aren't they already set in the context?

nicklasl (approved):

Confidence/src/main/java/com/spotify/confidence/AndroidLifecycleEventProducer.kt
fabriziodemaria: Good point, maybe we shouldn't add them. Will change and align Swift

nicklasl (dismissed):

Confidence/src/main/java/com/spotify/confidence/AndroidLifecycleEventProducer.kt
```diff
@@ -128,10 +128,10 @@ class AndroidLifecycleEventProducer(
...
         val currentBuild = ConfidenceValue.String(packageInfo?.getVersionCode().toString() ?: "")
 
-        // Get the previous recorded version.
-        val previousVersion = sharedPreferences
-            .getString(APP_VERSION, null)
-            ?.let(ConfidenceValue::String)
+        contextFlow.value =
+            mapOf(APP_VERSION_KEY to currentVersion)
+        contextFlow.value =
+            mapOf(APP_BUILD_KEY to currentBuild)
```
fabriziodemaria, line 134: contextFlow is of type `MutableStateFlow` that seems to lose some updates if we change value too quickly. I noticed app_build being set to the context, but not app_version.
I think it's better to fix this in this PR

vahidlazio: oh yes. the way we set the context flow here is wrong. mutableStateFlow is designed to only keep the last value. which is what we want, however in setting it, we need to change the whole value. in this case, we need to get the current value, which is a map, update the map entry and put back the whole map again into the contextFlow value. we can create a function to do all that and use it every where. the helper function can accept a key value to be updated in the value of context flow map.

Confidence/src/main/java/com/spotify/confidence/Confidence.kt
```diff
@@ -43,7 +44,7 @@ class Confidence internal constructor(
     private val contextChanges: Flow<Map<String, ConfidenceValue>> = contextMap
         .drop(1)
         .distinctUntilChanged()
-    private val coroutineScope = CoroutineScope(dispatcher)
+    private val coroutineScope = CoroutineScope(SupervisorJob() + dispatcher)
```
fabriziodemaria, line 47: Without this, network errors or missing client secret would cause this coroutine to crash and prevent many crucial operations relying on it

nicklasl: can we move this change out into a separate PR?

PR #165, opened 2024-06-06
feat: add DebugLogger to Confidence

nickybondarenko:


Confidence/src/main/java/com/spotify/confidence/DebugLogger.kt
```diff
@@ -0,0 +1,35 @@
...
+            DebugLoggerLevel.NONE -> {
+                // do nothing
+            }
+        }
+    }
+
+    internal fun logMessage(tag: String, message: String, isWarning: Boolean = false) {
+        if (!isWarning) {
+            when (level) {
+                DebugLoggerLevel.VERBOSE -> Log.v(tag, message)
```
vahidlazio, line 21: these cases seem all to be the same. could you explain the logic here please? what is `isWarning` and why that matters?

nickybondarenko: I think we can combine Verbose and Debug, since they both can handle logging exceptions. A warning, from what I understand, would show up differently in the Logcat, so I wanted to handle it differently - and right now you'd only get a warning if the storage is full.

Confidence/src/main/java/com/spotify/confidence/DebugLogger.kt
```diff
@@ -0,0 +1,47 @@
+package com.spotify.confidence
+
+import android.util.Log
+
+internal class DebugLogger {
+    var level: DebugLoggerLevel = DebugLoggerLevel.NONE
+
+    internal fun logEvent(tag: String, event: EngineEvent, details: String) {
+        when (level) {
```
vahidlazio, line 9: can we just create a private function like log(message: String) and not repeat this switch case for each log function?

Confidence/src/main/java/com/spotify/confidence/Confidence.kt
```diff
@@ -226,6 +237,7 @@ class Confidence internal constructor(
             eventSenderEngine.stop()
         }
         coroutineScope.cancel()
+        disableDebugLogger()
```
andreas-karlsson, line 240: Maybe we don't want to silence the logger explicitly? It should be come silent once everything is "closed" and if it doesn't that'd be valuable debugging info that something is broken.

nickybondarenko: Good point!

Confidence/src/main/java/com/spotify/confidence/Confidence.kt
```diff
@@ -74,6 +81,7 @@ class Confidence internal constructor(
     }
 
     fun apply(flagName: String, resolveToken: String) {
+        debugLogger?.logFlags("ApplyFlag", flagName)
```
vahidlazio, line 84: here we are calling the log function before the apply and in the context functions we call the logging after the operation. do we want to be consistent with the meaning of the logs we provide? do they mean we have called and done apply or resolve or it means we are about to start the operation?

Confidence/src/main/java/com/spotify/confidence/Confidence.kt
```diff
@@ -263,6 +282,7 @@ object ConfidenceFactory {
         val visitorId = ConfidenceValue.String(VisitorUtil.getId(context))
         val initContext = initialContext.toMutableMap()
         initContext[VISITOR_ID_CONTEXT_KEY] = visitorId
+        debugLogger?.let { debugLogger.logContext(initContext) }
```
vahidlazio, line 285: or `debugLogger?.logContext`?

Confidence/src/main/java/com/spotify/confidence/Confidence.kt
nickybondarenko: oh great point. i think we should call the logs after we do the operation, that would be consistent with what i had in mind. thanks for pointing it out!

Confidence/src/test/java/com/spotify/confidence/EventSenderIntegrationTest.kt
```diff
@@ -118,29 +118,34 @@ class EventSenderIntegrationTest {
...
+            flagResolver = mock(),
+            debugLogger = debugLogger
         )
         val eventSender = this@EventSenderIntegrationTest.eventSender
         val eventCount = 4 * batchSize + 2
         requireNotNull(eventSender)
         repeat(eventCount) {
             eventSender.track("navigate")
         }
+        Assert.assertEquals(eventCount * 4, debugLogger.eventsLogged)
```
vahidlazio, line 146: what is event count * 4?

Confidence/src/test/java/com/spotify/confidence/EventSenderIntegrationTest.kt
```diff
@@ -118,29 +118,34 @@ class EventSenderIntegrationTest {
...
         advanceUntilIdle()
+        Assert.assertEquals(18, debugLogger.messagesLogged)
```
vahidlazio, line 148: what is 18?

Confidence/src/test/java/com/spotify/confidence/EventSenderIntegrationTest.kt
nickybondarenko: so eventsLogged isn't a unique event, we log events when they get sent/stored/etc, so all in all we have 4 different times/places where we log events, thus *4.

Confidence/src/test/java/com/spotify/confidence/EventSenderIntegrationTest.kt
nickybondarenko: Same here, .messagesLogged isn't a unique log, but is triggered differently depending on the event. I was debating having it in the tests in general, since it's an arbitrary number, but decided to add and discuss if we want to keep it. I would like to keep it for the test purposes tbh. WDYT?

Confidence/src/test/java/com/spotify/confidence/EventSenderIntegrationTest.kt
vahidlazio: imo those etc are very important to know what we are testing. what if we change the policy numbers to one more but we have a bug that leads to not send another log message? then the count of number of logs is still 4 but the tests are passing which has to fail instead. so it's important to know what we are asserting. 
currently we log events for : emitting, writing to disk, flush policies hit (i'm also sceptical about logging this since it's not really related to any event lifecycle and it's pure implementation details, also the message is misleading since the flush policy is not really triggered for the event as it's stated in the message). is there more?

Confidence/src/test/java/com/spotify/confidence/EventSenderIntegrationTest.kt
vahidlazio: one way we could test this more deterministically was to introduce the type in logEvent, like, `logEvent(EventType.FlushTriggered ...)` or something like this, then we could test this better. I like to keep the test but somehow comment something or make it clear what this 18 means.

Confidence/src/test/java/com/spotify/confidence/EventSenderIntegrationTest.kt
vahidlazio: my suggestion is to at least comment where exactly these 4 logs are coming from.

nickybondarenko: sounds good! i had this comment previously, but wasn't sure if it should be in the tests. adding it back with elaboration

Confidence/src/test/java/com/spotify/confidence/EventSenderIntegrationTest.kt
nickybondarenko: I'll add a comment!

vahidlazio (approved):

PR #166, opened 2024-06-14
feat: Expose set sdk metadata

vahidlazio:


Confidence/src/main/java/com/spotify/confidence/Confidence.kt
```diff
@@ -57,6 +57,12 @@ class Confidence internal constructor(
         }
     }
 
+    fun setSdk(sdk: SdkMetadata) {
+        (flagResolver as RemoteFlagResolver).setSdk(sdk)
+        (eventSenderEngine as EventSenderEngineImpl).setSdk(sdk)
+        (flagApplierClient as FlagApplierClientImpl).setSdk(sdk)
+    }
```
nicklasl, line 64: Is it really not possible to only support passing it in the `create()` function?

Having it mutable isn't really something we want to support, right?

vahidlazio: yes but then what would be the default? if someone wants to use confidence in conjunction with open feature, do we really want to force them to pass the correct Id or additional parameter when creating confidence? or we want to do it automatically?
another way is to pass the `apiKey` and other confidence parameters to the provider creation function and then provider to create the confidence.

vahidlazio: we can make it immutable but take into account that users of open feature, in order to have the correct id, should pass `SDK_ID_KOTLIN_PROVIDER`, and then my worry is, it will be easy for them to just not do it and our telemetry will be incorrect.

nicklasl: Hmm... I didn't think about the OpenFeature provider case but purely for Flutter...

Can we explore different naming for create functions depending on the use case?

Or I wonder if we can pivot on the creation of the OF provider somehow to have it create the Confidence-instance behind the scenes or something? -- a bit strange since we likely want to also keep the reference to Confidence.

vahidlazio: > a bit strange since we likely want to also keep the reference to Confidence.
exactly!
for flutter, I already added in this PR the `sdk: SdkMetadata` parameter which can be passed in create functionn. `setSdk` is/will be called only within the provider context.

Provider/src/main/java/com/spotify/confidence/openfeature/ConfidenceFeatureProvider.kt
```diff
@@ -143,6 +148,23 @@ class ConfidenceFeatureProvider private constructor(
     companion object {
         private class ConfidenceMetadata(override var name: String? = "confidence") : ProviderMetadata
 
+        fun createConfidence(
+            context: Context,
+            clientSecret: String,
+            initialContext: Map<String, ConfidenceValue> = mapOf(),
+            region: ConfidenceRegion = ConfidenceRegion.GLOBAL,
+            dispatcher: CoroutineDispatcher = Dispatchers.IO
+        ): Confidence {
```
fabriziodemaria, line 157: I am wondering if we could introduce a new interface `ConfidenceForOpenFeature` or something, and make the `create` function below accept that, as a way to force users to use this `createConfidence`

fabriziodemaria (approved):
My comment is just an idea, disregard if you want to merge

Provider/src/main/java/com/spotify/confidence/openfeature/ConfidenceFeatureProvider.kt
vahidlazio: in this case `ConfidenceForOpenFeature ` should inherit both from event sender and from `Contextual`

vahidlazio: and all other functions are from confidence object itself, we don't have a confidence interface currently.

fabriziodemaria: Perhaps worth exploring, but I am also not strongly convinced these extra interfaces will actually help the user

vahidlazio: do you agree taking this in another PR and moving forward with this PR?

PR #172, opened 2024-06-19
fix: Fix the bug regarding accessing flags using with context

vahidlazio:


Confidence/src/main/java/com/spotify/confidence/Confidence.kt
```diff
@@ -144,7 +146,7 @@ class Confidence internal constructor(
             it.getContext().filterKeys { key -> !removedKeys.contains(key) } + contextMap.value
         } ?: contextMap.value
 
-    override fun withContext(context: Map<String, ConfidenceValue>): Confidence = Confidence(
+    override fun withContext(context: Map<String, ConfidenceValue>): ConfidenceAPI = Confidence(
```
fabriziodemaria, line 149: The `EventSender` interface already extends `Contextual`: can we just return the `EventSender` type from withContext?

vahidlazio: I think we can good point.

nickybondarenko (approved):

nicklasl (approved):

fabriziodemaria (approved):

PR #174, opened 2024-06-26
fix: update how debug logging works

nicklasl:
Only logged invocation coming from the SDK that matches the logging level defined by the customer will be printed to Logcat.

The idea here is to have the customer supply the lowest level that they want to get logs for -- they would do that in the `create` function where we now default to `LoggingLevel.WARN`.

The SDK is then opinionated about which level to log different things on, for example:
```kotlin
    override fun logEvent(action: String, event: EngineEvent) {
        debug("[$action] $event")
    }
    // and
    override fun logContext(context: Map<String, ConfidenceValue>) {
        verbose(context.toString())
    }
```

Warnings and errors can contain message from a `throwable`.


Confidence/src/main/java/com/spotify/confidence/Confidence.kt
```diff
@@ -66,9 +66,7 @@ class Confidence internal constructor(
 
     private suspend fun resolve(flags: List<String>): Result<FlagResolution> {
         debugLogger?.let {
-            for (flag in flags) {
-                debugLogger.logFlags("ResolveFlag", flag)
-            }
+            debugLogger.logFlag("Resolve")
```
nickybondarenko, line 69: Do we want to explain why first resolve will be empty? Or is this more of a future fix/refactor territory?

nicklasl: I'm thinking we can provide a new PR where we clean up the API (`resolve` functions should not take the list of flags).

nickybondarenko (dismissed):

nicklasl:
Note to self: Don't forget to also update the Readme docs.

PR #182, opened 2024-11-13
fix: guard possible changes in serialization

nickybondarenko:
This PR fixes app crash after breaking changes to how we serialize storing flag and apply data. If the app can't read the storage (format can't be read), it wipes the file and fetches latest flags and applies in case of sticky assignment.

### DIFF
--- a/Confidence/src/test/java/com/spotify/confidence/FileDiskStorageTest.kt
+++ b/Confidence/src/test/java/com/spotify/confidence/FileDiskStorageTest.kt
@@ -1,0 +1,68 @@
+package com.spotify.confidence
+
+import com.spotify.confidence.apply.ApplyInstance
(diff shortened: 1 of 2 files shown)


Confidence/src/test/java/com/spotify/confidence/FileDiskStorageTest.kt
```diff
@@ -0,0 +1,68 @@
+package com.spotify.confidence
+
+import com.spotify.confidence.apply.ApplyInstance
+import com.spotify.confidence.apply.EventStatus
+import com.spotify.confidence.cache.FileDiskStorage
+import com.spotify.confidence.client.ConfidenceValueMap
```
nicklasl, line 6: unused?

nicklasl (approved):

PR #183, opened 2024-11-13
feat: Resolve Debug to redirect to UI tool on web

nickybondarenko:
This PR adds logging to DEBUG layer to redirect user to our web UI for Resolve debugging in case of anything but success resolve (MATCH).

### DIFF
--- a/Confidence/src/main/java/com/spotify/confidence/Confidence.kt
+++ b/Confidence/src/main/java/com/spotify/confidence/Confidence.kt
@@ -106,15 +106,24 @@ class Confidence internal constructor(
     fun <T> getFlag(
         key: String,
         default: T
-    ): Evaluation<T> = cache.get().getEvaluation(
-        key,
-        default,
-        getContext()
-    ) { flagName, resolveToken ->
-        // this lambda will be invoked inside the evaluation process
-        // and only if the resolve reason is not targeting key error.
-        apply(flagName, resolveToken)
+    ): Evaluation<T> {
+        val eval = cache.get().getEvaluation(
+            key,
+            default,
+            getContext()
+        ) { flagName, resolveToken ->
+            // this lambda will be invoked inside the evaluation process
+            // and only if the resolve reason is not targeting key error.
+            apply(flagName, resolveToken)
+        }
+        if (eval.reason !== ResolveReason.RESOLVE_REASON_MATCH) {
+            val context = this.getContext()
+            val flag = key.splitToSequence(".").first()
+            debugLogger?.logResolve(flag, context)
+        }
+        return eval
     }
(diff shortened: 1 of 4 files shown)


Confidence/src/main/java/com/spotify/confidence/DebugLogger.kt
```diff
@@ -35,6 +36,11 @@ internal class DebugLoggerImpl(private val filterLevel: LoggingLevel) : DebugLog
         verbose("[$action] $context")
     }
 
+    override fun logResolve(flag: String, context: Map<String, ConfidenceValue>) {
+        val trimContext = "$context".replace(" ", "")
+        debug("[Resolve Debug] https://app.confidence.spotify.com/flags/resolver-test?flag=flags/$flag&context=$trimContext")
```
nickybondarenko, line 41: Currently, the context doesn't show up in the UI.

Confidence/src/main/java/com/spotify/confidence/DebugLogger.kt
```diff
@@ -35,6 +36,11 @@ internal class DebugLoggerImpl(private val filterLevel: LoggingLevel) : DebugLog
         verbose("[$action] $context")
     }
 
+    override fun logResolve(flag: String, context: Map<String, ConfidenceValue>) {
+        val trimContext = "$context".replace(" ", "")
```
nicklasl, line 40: not sure you want to do this. 
first; think about what format is expected. probably some json?
also, if we trim context values from `" "` then any context key or value that contains a `" "` would be incorrect, right?

Confidence/src/main/java/com/spotify/confidence/Confidence.kt
```diff
@@ -106,16 +106,25 @@ class Confidence internal constructor(
...
+        val eval = cache.get().getEvaluation(
+            key,
+            default,
+            getContext()
+        ) { flagName, resolveToken ->
+            // this lambda will be invoked inside the evaluation process
+            // and only if the resolve reason is not targeting key error.
+            apply(flagName, resolveToken)
+        }
+        if (eval.reason !== ResolveReason.RESOLVE_REASON_MATCH) {
```
nicklasl, line 119: I think we should do it even if we get a match.

Confidence/src/main/java/com/spotify/confidence/DebugLogger.kt
nickybondarenko: Right, I will figure out how to translate the context into a JSON..

nicklasl (dismissed):

nicklasl (approved):

PR #185, opened 2024-11-22
fix: awaitReconcile race with contextChanges

nicklasl:
There was a case for race conditions that could occur when calling `putContext()` followed by a `awaitReconciliation()` where the `contextChanges` updates did not trigger a `fetchAndActivate` before `awaitReconciliation` had finished.

### DIFF
--- a/Confidence/api/Confidence.api
+++ b/Confidence/api/Confidence.api
@@ -33,7 +33,8 @@ public final class com/spotify/confidence/Confidence : com/spotify/confidence/Co
 	public final fun activate ()V
 	public final fun apply (Ljava/lang/String;Ljava/lang/String;)V
 	public final fun asyncFetch ()V
-	public final fun awaitReconciliation (Lkotlin/coroutines/Continuation;)Ljava/lang/Object;
+	public final fun awaitReconciliation (JLkotlin/coroutines/Continuation;)Ljava/lang/Object;
+	public static synthetic fun awaitReconciliation$default (Lcom/spotify/confidence/Confidence;JLkotlin/coroutines/Continuation;ILjava/lang/Object;)Ljava/lang/Object;
 	public final fun fetchAndActivate (Lkotlin/coroutines/Continuation;)Ljava/lang/Object;
 	public fun flush ()V
 	public fun getContext ()Ljava/util/Map;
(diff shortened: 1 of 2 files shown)


Confidence/api/Confidence.api
```diff
@@ -102,8 +103,8 @@ public final class com/spotify/confidence/ConfidenceError$ParseError : java/lang
...
-	public final fun create (Landroid/content/Context;Ljava/lang/String;Ljava/util/Map;Lcom/spotify/confidence/ConfidenceRegion;Lkotlinx/coroutines/CoroutineDispatcher;Lcom/spotify/confidence/LoggingLevel;)Lcom/spotify/confidence/Confidence;
-	public static synthetic fun create$default (Lcom/spotify/confidence/ConfidenceFactory;Landroid/content/Context;Ljava/lang/String;Ljava/util/Map;Lcom/spotify/confidence/ConfidenceRegion;Lkotlinx/coroutines/CoroutineDispatcher;Lcom/spotify/confidence/LoggingLevel;ILjava/lang/Object;)Lcom/spotify/confidence/Confidence;
+	public final fun create (Landroid/content/Context;Ljava/lang/String;Ljava/util/Map;Lcom/spotify/confidence/ConfidenceRegion;Lkotlinx/coroutines/CoroutineDispatcher;Lcom/spotify/confidence/LoggingLevel;J)Lcom/spotify/confidence/Confidence;
+	public static synthetic fun create$default (Lcom/spotify/confidence/ConfidenceFactory;Landroid/content/Context;Ljava/lang/String;Ljava/util/Map;Lcom/spotify/confidence/ConfidenceRegion;Lkotlinx/coroutines/CoroutineDispatcher;Lcom/spotify/confidence/LoggingLevel;JILjava/lang/Object;)Lcom/spotify/confidence/Confidence;
```
nickybondarenko, line 107: Why did this part of the API change? 🤔

Confidence/src/main/java/com/spotify/confidence/Confidence.kt
```diff
@@ -77,10 +80,23 @@ class Confidence internal constructor(
...
-            activate()
+    suspend fun awaitReconciliation(timeoutMillis: Long = 5000) {
+        try {
+            withTimeout(timeoutMillis) {
+                while (currentFetchJob == null ||
+                    currentFetchJob?.isCompleted == true ||
+                    currentFetchJob?.isCancelled == true
+                ) {
+                    // If this function is called just after a putContext it will likely just delay once.
+                    delay(1)
```
fabriziodemaria, line 91: I think the un-predictability of when `putContext` is called + the short timespan of `1ms` _could_ make this code quite inefficient

Confidence/src/main/java/com/spotify/confidence/Confidence.kt
```diff
@@ -77,10 +80,23 @@ class Confidence internal constructor(
...
-    suspend fun awaitReconciliation() {
-        if (currentFetchJob != null) {
-            currentFetchJob?.join()
-            activate()
+    suspend fun awaitReconciliation(timeoutMillis: Long = 5000) {
+        try {
+            withTimeout(timeoutMillis) {
+                while (currentFetchJob == null ||
+                    currentFetchJob?.isCompleted == true ||
+                    currentFetchJob?.isCancelled == true
```
fabriziodemaria, line 88: Maybe replace these with `currentFetchJob?.isActive != true`

Confidence/src/main/java/com/spotify/confidence/Confidence.kt
```diff
@@ -77,10 +80,23 @@ class Confidence internal constructor(
         return flagResolver.resolve(flags, getContext())
     }
 
-    suspend fun awaitReconciliation() {
-        if (currentFetchJob != null) {
-            currentFetchJob?.join()
-            activate()
+    suspend fun awaitReconciliation(timeoutMillis: Long = 5000) {
```
fabriziodemaria, line 83: Does this support multiple concurrent calls? My concern is:
`awaitReconciliation (1)`
`awaitReconciliation (2)`
`putContext` -> this unlocks only (1)

fabriziodemaria (commented):
Added a few comments about various corner cases that I would like to discuss a bit further

Confidence/src/main/java/com/spotify/confidence/Confidence.kt
nicklasl: Mmm, perhaps. However I'm not certain having this is a great common timeout is great (imagine you call this function when we are actually NOT doing a network call, then it'll always just wait for timeoutMillis)... I might push an update.

Another approach would be to "delay" until `cache.flagEvaluation.context == getContext()` (but diffing them in some hashy way of course).

Confidence/src/main/java/com/spotify/confidence/Confidence.kt
nicklasl: Please explain more, not sure I understand the concern.

fabriziodemaria: I think what I am getting at is: are there more idiomatic ways to achieve this behaviour? I am looking at `job.invokeOnCompletion`, for example, but I don't know if it does exactly what we need

PR #187, opened 2024-11-28
fix: await reconciliation bug

nicklasl:
There was a case for race conditions that could occur when calling putContext() followed by a awaitReconciliation() where the contextChanges updates did not trigger a fetchAndActivate before awaitReconciliation had finished.

Replaces https://github.com/spotify/confidence-sdk-android/pull/185

### DIFF
--- a/Confidence/src/main/java/com/spotify/confidence/Confidence.kt
+++ b/Confidence/src/main/java/com/spotify/confidence/Confidence.kt
@@ -15,11 +15,11 @@ import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.Job
 import kotlinx.coroutines.SupervisorJob
 import kotlinx.coroutines.TimeoutCancellationException
-import kotlinx.coroutines.async
 import kotlinx.coroutines.cancel
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.withTimeout
+import kotlinx.coroutines.yield
 import kotlinx.serialization.Serializable
 import kotlinx.serialization.json.Json
 import kotlinx.serialization.json.encodeToJsonElement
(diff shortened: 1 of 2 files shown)


Confidence/src/main/java/com/spotify/confidence/Confidence.kt
```diff
@@ -60,11 +63,33 @@ class Confidence internal constructor(
...
-    suspend fun awaitReconciliation() {
-        if (currentFetchJob != null) {
+    suspend fun awaitReconciliation(timeoutMillis: Long = 5000) {
+        val delayTime = 3L
+        if (timeoutMillis <= delayTime) error("timeoutMillis need to be larger than $delayTime")
+        debugLogger?.logMessage("reconciliation started")
+        withSafeTimeout(delayTime) {
+            while (currentFetchJob == null) {
+                delay(1)
+                debugLogger?.logMessage("delayed")
```
nickybondarenko, line 73: Why do we need to delay if there's currently no fetch jobs scheduled?

nicklasl: Good comment and this made us take a closer look together and come up with a better solution.

Confidence/src/main/java/com/spotify/confidence/Confidence.kt
```diff
@@ -60,11 +63,25 @@ class Confidence internal constructor(
         }
     }
 
-    suspend fun awaitReconciliation() {
-        if (currentFetchJob != null) {
+    suspend fun awaitReconciliation(timeoutMillis: Long = 5000) {
+        if (timeoutMillis <= 0) error("timeoutMillis need to be larger than 0")
+        debugLogger?.logMessage("reconciliation started")
+        yield() // will make sure that we respect other coroutine scopes triggered before this
```
nicklasl, line 69: This line is the actual fix. The rest is mainly just improvements.

nickybondarenko (approved):

PR #191, opened 2025-01-27
feat!: refactor context provider

nicklasl:
The main work here is to stop exposing the `AndroidLifecycleEventProducer` and to restructure how we expose the ?`Producer` interface.
Keeping multiple breaking changes in one PR.

### DIFF
--- a/Confidence/src/main/java/com/spotify/confidence/Confidence.kt
+++ b/Confidence/src/main/java/com/spotify/confidence/Confidence.kt
@@ -46,7 +46,7 @@ class Confidence internal constructor(
     private var currentFetchJob: Job? = null
 
     private val coroutineScope = CoroutineScope(SupervisorJob() + dispatcher)
-    private val eventProducers: MutableList<EventProducer> = mutableListOf()
+    private val producers: MutableList<Producer> = mutableListOf()
 
     private val flagApplier = FlagApplierWithRetries(
         client = flagApplierClient,
@@ -276,31 +276,34 @@ class Confidence internal constructor(
         activate()
     }
 
-    override fun track(eventProducer: EventProducer) {
-        coroutineScope.launch {
-            eventProducer
-                .events()
-                .collect { event ->
-                    eventSenderEngine.emit(
-                        event.name,
-                        event.data,
-                        getContext()
-                    )
-                    if (event.shouldFlush) {
-                        eventSenderEngine.flush()
+    override fun track(producer: Producer) {
+        (producer as? EventProducer)?.let {
+            coroutineScope.launch {
+                it
+                    .events()
+                    .collect { event ->
+                        eventSenderEngine.emit(
+                            event.name,
+                            event.data,
+                            getContext()
+                        )
+                        if (event.shouldFlush) {
+                            eventSenderEngine.flush()
+                        }
                     }
-                }
+            }
         }
-
-        coroutineScope.launch {
-            eventProducer.contextChanges()
-                .collect(this@Confidence::putContext)
+        (producer as? ContextProducer)?.let {
+            coroutineScope.launch {
+                it.contextChanges()
+                    .collect(this@Confidence::putContext)
+            }
         }
-        eventProducers.add(eventProducer)
+        producers.add(producer)
     }
 
     override fun stop() {
-        for (producer in eventProducers) {
+        for (producer in producers) {
             producer.stop()
         }
         if (parent == null) {
--- a/Confidence/src/main/java/com/spotify/confidence/Confidence.kt
+++ b/Confidence/src/main/java/com/spotify/confidence/Confidence.kt
@@ -277,29 +277,24 @@ class Confidence internal constructor(
     }
 
     override fun track(producer: Producer) {
-        (producer as? EventProducer)?.let {
-            coroutineScope.launch {
-                it
-                    .events()
-                    .collect { event ->
+        coroutineScope.launch {
+            producer.updates().collect { update ->
+                when (update) {
+                    is Update.Event -> {
                         eventSenderEngine.emit(
-                            event.name,
-                            event.data,
+                            update.name,
+                            update.data,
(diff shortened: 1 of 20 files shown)


Confidence/src/main/java/com/spotify/confidence/ConfidenceDeviceInfoContextProducer.kt
```diff
@@ -0,0 +1,91 @@
...
+        if (withBundleId) {
+            val bundleId = ConfidenceValue.String(applicationContext.packageName)
+            context += mapOf(BUNDLE_ID_CONTEXT_KEY to bundleId)
+        }
+        if (withDeviceInfo) {
+            val deviceInfo = mapOf(
+                SYSTEM_NAME_CONTEXT_KEY to ConfidenceValue.String("Android"),
+                DEVICE_BRAND_CONTEXT_KEY to ConfidenceValue.String(Build.BRAND),
+                DEVICE_MODEL_CONTEXT_KEY to ConfidenceValue.String(Build.MODEL),
+                SYSTEM_VERSION_CONTEXT_KEY to ConfidenceValue.Double(Build.VERSION.SDK_INT.toDouble())
```
nickybondarenko, line 48: Could we add some comments here to document examples?

Confidence/src/main/java/com/spotify/confidence/ConfidenceDeviceInfoContextProducer.kt
```diff
@@ -0,0 +1,91 @@
...
+    withBundleId: Boolean = false,
+    withDeviceInfo: Boolean = false,
+    withLocale: Boolean = false,
+) : ContextProducer {
+    private val contextFlow = MutableStateFlow<Map<String, ConfidenceValue>>(mapOf())
+    private val packageInfo: PackageInfo? = try {
+        @Suppress("DEPRECATION")
+        applicationContext.packageManager.getPackageInfo(applicationContext.packageName, 0)
+    } catch (e: PackageManager.NameNotFoundException) {
+        Log.w(DebugLogger.TAG, "Failed to get package info", e)
```
nickybondarenko, line 24: nit: Should we add Confidence logger here or keep the system one?

nicklasl: I agree it would have been nice but this class (`ConfidenceDeviceInfoContextProducer`) is instantiated from outside the SDK (which we want) while the DebugLogger is internal so we can't actually use it.

nickybondarenko (dismissed):

Confidence/src/main/java/com/spotify/confidence/Producer.kt
```diff
@@ -8,9 +8,14 @@ data class Event(
     val shouldFlush: Boolean = false
 )
 
-interface EventProducer {
+sealed interface Producer {
```
andreas-karlsson, line 11: If a Producer can produce both events and context changes, I think it's a requirement that the implementor can control the exact order of these. Not sure if that's possible with the proposed design. Maybe an alternative would be to use one flow of Event _or_ ContextChange. Either way I think we should have a test to verify order can be controlled.

Confidence/src/main/java/com/spotify/confidence/ConfidenceDeviceInfoContextProducer.kt
```diff
@@ -0,0 +1,91 @@
...
+    companion object {
+        const val APP_VERSION_CONTEXT_KEY = "app_version"
+        const val APP_BUILD_CONTEXT_KEY = "app_build"
+        const val BUNDLE_ID_CONTEXT_KEY = "bundle_id"
+        const val SYSTEM_NAME_CONTEXT_KEY = "system_name"
+        const val DEVICE_BRAND_CONTEXT_KEY = "device_brand"
+        const val DEVICE_MODEL_CONTEXT_KEY = "device_model"
+        const val SYSTEM_VERSION_CONTEXT_KEY = "system_version"
+        const val LOCAL_IDENTIFIER_CONTEXT_KEY = "locale_identifier"
+        const val PREFERRED_LANGUAGES_CONTEXT_KEY = "preferred_languages"
```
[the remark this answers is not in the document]
nicklasl: Pushed some changes to align with this spec as much as I could.

Confidence/src/main/java/com/spotify/confidence/Producer.kt
```diff
@@ -0,0 +1,22 @@
+package com.spotify.confidence
+
+import kotlinx.coroutines.flow.Flow
+
+sealed interface Update {
```
andreas-karlsson, line 5: Probably doesn't matter but I checked an article on algebraic data types in Kotlin (https://medium.com/sharenowtech/kotlin-adt-74472319962a) and they use `sealed class` what's the difference of using class/interface here?

Confidence/src/main/java/com/spotify/confidence/Confidence.kt
```diff
@@ -276,31 +276,29 @@ class Confidence internal constructor(
...
+                            update.name,
+                            update.data,
+                            getContext()
+                        )
+                        if (update.shouldFlush) {
+                            eventSenderEngine.flush()
+                        }
                     }
+                    is Update.ContextUpdate -> putContext(update.context)
                 }
```
andreas-karlsson, line 294: Nice! I like this change of having them in one flow.

andreas-karlsson (approved):

Confidence/src/main/java/com/spotify/confidence/Producer.kt
nicklasl: I have never understood the difference. It MAY be more technically correct to use a `sealed class` in this case but probably a `sealed interface` for the `Producer`(when it was still sealed).

nickybondarenko (approved):

PR #192, opened 2025-02-04
feat: add shouldApply to the apply call to depend on backend

nickybondarenko:
The SDK will use `shouldApply` boolean coming from backend service when sending an apply event.

### DIFF
--- a/Confidence/src/main/java/com/spotify/confidence/Confidence.kt
+++ b/Confidence/src/main/java/com/spotify/confidence/Confidence.kt
@@ -88,10 +88,9 @@ class Confidence internal constructor(
      * Apply a flag
      * @param flagName name of the flag.
      * @param resolveToken resolve token.
-     * @param shouldApply whether to send an apply event or not
      */
-    fun apply(flagName: String, resolveToken: String, shouldApply: Boolean) {
-        flagApplier.apply(flagName, resolveToken, shouldApply)
+    fun apply(flagName: String, resolveToken: String) {
+        flagApplier.apply(flagName, resolveToken)
         debugLogger?.logFlag("Apply", flagName)
     }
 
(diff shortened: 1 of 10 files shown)


Confidence/src/main/java/com/spotify/confidence/Confidence.kt
```diff
@@ -115,10 +116,10 @@ class Confidence internal constructor(
             key,
             default,
             evaluationContext
-        ) { flagName, resolveToken ->
+        ) { flagName, resolveToken, shouldApply ->
             // this lambda will be invoked inside the evaluation process
             // and only if the resolve reason is not targeting key error.
-            apply(flagName, resolveToken)
+            apply(flagName, resolveToken, shouldApply)
```
nicklasl, line 122: did you consider adding the `if (shouldApply)` around this place instead of deeper down in the call stack?

Confidence/src/main/java/com/spotify/confidence/ConfidenceFlagEvaluation.kt
```diff
@@ -25,7 +25,7 @@ fun <T> FlagResolution?.getEvaluation(
         )
 
     if (resolvedFlag.reason != ResolveReason.RESOLVE_REASON_TARGETING_KEY_ERROR) {
-        applyFlag(parsedKey.flagName, resolveToken)
+        applyFlag(parsedKey.flagName, resolveToken, resolvedFlag.shouldApply)
```
nicklasl, line 28: Or maybe you could even have the `if(resolvedFlag.shouldApply)` here?

nickybondarenko: Good point, this place seems like a good check. Will also simplify the change. Will also be the same as what we have in the js PR I believe

nicklasl (approved):
LGTM 🍬

PR #195, opened 2025-02-11
chore: update OF version

nickybondarenko:


Confidence/build.gradle.kts
```diff
@@ -12,7 +12,7 @@ plugins {
 val providerVersion = project.extra["version"].toString()
 
 object Versions {
-    const val openFeatureSDK = "0.2.3"
+    const val openFeatureSDK = "0.3.3"
```
nicklasl, line 15: why is this even here?

nicklasl (approved):

PR #197, opened 2025-03-07
feat(Provider)!: update to OF SDK 0.4.1

nicklasl:
This is only breaking for the Provider but not for the underlying SDK.


Confidence/src/main/java/com/spotify/confidence/ConfidenceContext.kt
```diff
@@ -19,19 +19,22 @@ interface Contextual : ConfidenceContextProvider {
     /**
      * Add entry to context
      * @param context context to add.
+     * @param skipFetch if true, the context will not initiate a new fetch.
      */
-    fun putContext(context: Map<String, ConfidenceValue>)
```
fabriziodemaria, line 23: Should we try to align the api structure with the ios one? https://github.com/spotify/confidence-sdk-swift/blob/a4b1e0af5ccfc83d61ca566a78f4a4181b112d63/Sources/Confidence/Confidence.swift#L220

Confidence/src/main/java/com/spotify/confidence/Confidence.kt
```diff
@@ -132,20 +132,20 @@ class Confidence internal constructor(
     }
 
     @Synchronized
-    override fun putContext(key: String, value: ConfidenceValue) {
```
fabriziodemaria, line 135: Should we try to align the api structure with the ios one? https://github.com/spotify/confidence-sdk-swift/blob/a4b1e0af5ccfc83d61ca566a78f4a4181b112d63/Sources/Confidence/Confidence.swift#L220

nicklasl: I only implemented the `putContextLocal` with the args that were required for the specific integration. I don't know if I think they need to align. It is just an API to support our provider, if I could have hidden it in a nice way I would.

fabriziodemaria (approved):
LGTM

PR #210, opened 2025-09-01
docs: openfeature first focused docs

nicklasl:


README.md
```diff
@@ -20,19 +20,107 @@ Functionalities:
...
+    clientSecret = "<MY_SECRET>",
+    region = ConfidenceRegion.EUROPE,
+    loggingLevel = LoggingLevel.VERBOSE
+),  initialisationStrategy = InitialisationStrategy.FetchAndActivate)
+
+```
+Where `MY_SECRET` is an API key that can be generated in the [Confidence UI](https://confidence.spotify.com/console).
+The `loggingLevel` sets the verbosity level for logging to console. This can be useful while testing your integration with the Confidence SDK.
+
+_Note: the Confidence SDK has been intended to work as a single instance in your Application. Creating multiple instances in the same runtime could lead to unexpected behaviours._
```
fabriziodemaria, line 50: Maybe this note is not as relevant now that OpenFeature is used in this section

README.md
```diff
@@ -20,19 +20,107 @@ Functionalities:
...
+> [!IMPORTANT]
+> You should note that all feature flags in Confidence are complex objects implemented by the `ConfidenceValue.Struct` class but fear not! 
+> You can use a dot-notation to access properties within your flag object for ease of use.
+> Above we show how to access a string property from a flag. In that example the actual flag object is called `flag-name` and we access the string property `message` using the dot notation: `flag-name.message`.
+
+### Initialization strategy
+`initialisationStrategy` is a way to decide how Confidence should act when the provider is being set.
+- `ActivateAndFetchAsync` will make the last fetched flags available immediately and then trigger a background fetch to update the flags for future sessions.
+- `FetchAndActivate` will fetch the flags from the Confidence backend, stores the result on disk, and make the same data ready for the Application to be consumed.
+  If freshness of flags are the most important aspect for your use case, this is the strategy to go for. However, this will introduce a delay in making flags available to the application.
```
fabriziodemaria, line 104: This could be confusing, we might want to spend a few extra paragraph to show (maybe in code?) how to wait for a Provider to be ready and when it's safe to start evaluating in the different strategy scenarios

README.md
```diff
@@ -84,11 +172,6 @@ val messageValue = messageFlag.value
 // message and messageValue are the same
 ```
 
-> [!IMPORTANT]
-> You should note that all feature flags in Confidence are complex objects implemented by the `ConfidenceValue.Struct` class but fear not! 
-> You can use a dot-notation to access properties within your flag object for ease of use.
-> Above we show how to access a string property from a flag. In that example the actual flag object is called `flag-name` and we access the string property `message` using the dot notation: `flag-name.message`.
-
 ### Tracking an event
 Events are defined by a `name` and a `message` where the message is a key-value map of type `<String, ConfidenceValue>`. You can track an event using the `track` method.
```
fabriziodemaria, line 176: I wonder if we should remove this section while this feature is not yet available for production on the backend-side

README.md
```diff
@@ -20,19 +20,107 @@ Functionalities:
...
+
+### Initialization strategy
+`initialisationStrategy` is a way to decide how Confidence should act when the provider is being set.
+- `ActivateAndFetchAsync` will make the last fetched flags available immediately and then trigger a background fetch to update the flags for future sessions.
+- `FetchAndActivate` will fetch the flags from the Confidence backend, stores the result on disk, and make the same data ready for the Application to be consumed.
+  If freshness of flags are the most important aspect for your use case, this is the strategy to go for. However, this will introduce a delay in making flags available to the application.
+
+
+## Using the Confidence Standalone SDK
+If you want to use the Confidence SDK standalone, you can.
```
fabriziodemaria, line 108: Perhaps a few more words on what the standalone counterpart is and why we recommend using OF?

fabriziodemaria (commented):
Only README comments, most of them just suggestions

PR #219, opened 2025-09-15
fix!: no more buildconfig

nicklasl:


ConfidenceDemoApp/src/main/AndroidManifest.xml
```diff
@@ -10,6 +10,10 @@
             android:roundIcon="@mipmap/ic_launcher_round"
             android:supportsRtl="true"
             android:theme="@style/Theme.Provider">
+        
+        <meta-data
```
vahidlazio, line 14: should we add this in the readme?

vahidlazio (approved):

PR #252, opened 2026-08-07
feat: OpenFeature track and event delivery reliability

fabriziodemaria:
## Summary
- Align OpenFeature `track()` with Confidence event conventions (details → data, context merge, `"value"` / `"context"` overrides).
- Improve event delivery reliability: optional interval flush, startup flush, stop/shutdown flush, provider `shutdown()` → `flush()`.

## Test plan
- [x] Provider track tests
- [x] `EventSenderEngineReliabilityTest`
- [ ] Full CI on combined branch

Supersedes #253 (reliability) — both changes are on `openfeature-track-context`.

### DIFF
--- a/Confidence/src/main/java/com/spotify/confidence/Confidence.kt
+++ b/Confidence/src/main/java/com/spotify/confidence/Confidence.kt
@@ -385,6 +385,7 @@ object ConfidenceFactory {
      * @param loggingLevel allows to print warnings or debugging information to the local console.
      * @param timeoutMillis sets a timeout for completing an HTTP call. Defaults to 10 seconds
      * @param visitorIdContextKey key to use for the visitor id in the context. Defaults to "visitor_id".
+     * @param eventFlushIntervalMillis optional periodic flush interval in milliseconds. Disabled by default.
      */
     fun create(
         context: Context,
@@ -394,7 +395,8 @@ object ConfidenceFactory {
         dispatcher: CoroutineDispatcher = Dispatchers.IO,
         loggingLevel: LoggingLevel = LoggingLevel.WARN,
         timeoutMillis: Long = 10000,
-        visitorIdContextKey: String = VISITOR_ID_CONTEXT_KEY
+        visitorIdContextKey: String = VISITOR_ID_CONTEXT_KEY,
+        eventFlushIntervalMillis: Long? = null
     ): Confidence = create(
         context = context,
         clientSecret = clientSecret,
@@ -404,7 +406,8 @@ object ConfidenceFactory {
         loggingLevel = loggingLevel,
         timeoutMillis = timeoutMillis,
         visitorIdContextKey = visitorIdContextKey,
-        resolveBaseUrl = null
+        resolveBaseUrl = null,
+        eventFlushIntervalMillis = eventFlushIntervalMillis
     )
 
     /**
@@ -421,7 +424,8 @@ object ConfidenceFactory {
         dispatcher: CoroutineDispatcher = Dispatchers.IO,
         loggingLevel: LoggingLevel = LoggingLevel.WARN,
         timeoutMillis: Long = 10000,
-        visitorIdContextKey: String = VISITOR_ID_CONTEXT_KEY
+        visitorIdContextKey: String = VISITOR_ID_CONTEXT_KEY,
+        eventFlushIntervalMillis: Long? = null
     ): Confidence = create(
         context = context,
         clientSecret = clientSecret,
@@ -431,7 +435,8 @@ object ConfidenceFactory {
         loggingLevel = loggingLevel,
         timeoutMillis = timeoutMillis,
         visitorIdContextKey = visitorIdContextKey,
-        resolveBaseUrl = getResolveBaseUrl(region, resolveBaseUrl)
+        resolveBaseUrl = getResolveBaseUrl(region, resolveBaseUrl),
+        eventFlushIntervalMillis = eventFlushIntervalMillis
     )
 
     private fun create(
@@ -443,7 +448,8 @@ object ConfidenceFactory {
         loggingLevel: LoggingLevel,
         timeoutMillis: Long,
         visitorIdContextKey: String,
-        resolveBaseUrl: HttpUrl?
+        resolveBaseUrl: HttpUrl?,
+        eventFlushIntervalMillis: Long? = null
     ): Confidence {
         val debugLogger: DebugLogger? = if (loggingLevel == LoggingLevel.NONE) {
             null
@@ -458,7 +464,8 @@ object ConfidenceFactory {
             flushPolicies = listOf(minBatchSizeFlushPolicy),
             sdkMetadata = sdkMetadata,
             dispatcher = dispatcher,
-            debugLogger = debugLogger
+            debugLogger = debugLogger,
+            flushIntervalMillis = eventFlushIntervalMillis
         )
         val flagApplierClient = FlagApplierClientImpl(
             clientSecret,
--- a/Confidence/src/main/java/com/spotify/confidence/Confidence.kt
+++ b/Confidence/src/main/java/com/spotify/confidence/Confidence.kt
@@ -451,6 +451,9 @@ object ConfidenceFactory {
         resolveBaseUrl: HttpUrl?,
         eventFlushIntervalMillis: Long? = null
     ): Confidence {
+        require(eventFlushIntervalMillis == null || eventFlushIntervalMillis > 0) {
+            "eventFlushIntervalMillis must be positive, or null to disable periodic flushing"
+        }
         val debugLogger: DebugLogger? = if (loggingLevel == LoggingLevel.NONE) {
             null
         } else {
--- a/Confidence/src/main/java/com/spotify/confidence/Confidence.kt
+++ b/Confidence/src/main/java/com/spotify/confidence/Confidence.kt
@@ -254,7 +254,15 @@ class Confidence internal constructor(
         eventName: String,
         data: ConfidenceFieldsType
     ) {
-        eventSenderEngine.emit(eventName, data, getContext())
+        track(eventName, data, getContext())
+    }
+
+    override fun track(
+        eventName: String,
+        data: ConfidenceFieldsType,
+        eventContext: Map<String, ConfidenceValue>
+    ) {
+        eventSenderEngine.emit(eventName, data, eventContext)
     }
 
     override fun flush() {
--- a/Confidence/src/main/java/com/spotify/confidence/EventSender.kt
+++ b/Confidence/src/main/java/com/spotify/confidence/EventSender.kt
@@ -12,6 +12,16 @@ interface EventSender : Contextual {
     )
 
     /**
+     * Store a custom event to be tracked with an explicit evaluation context.
+     * @param eventContext evaluation context for this event only; does not mutate session context.
+     */
+    fun track(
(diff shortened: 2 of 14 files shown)


Confidence/src/main/java/com/spotify/confidence/EventSenderEngine.kt
```diff
@@ -30,11 +38,18 @@ internal class EventSenderEngineImpl(
...
     private val dispatcher: CoroutineDispatcher = Dispatchers.IO,
     private val sdkMetadata: SdkMetadata,
-    private val debugLogger: DebugLogger?
+    private val debugLogger: DebugLogger?,
+    private val flushIntervalMillis: Long? = null
 ) : EventSenderEngine {
-    private val writeReqChannel: Channel<EngineEvent> = Channel()
+    // Buffered so emit()/flush() can enqueue synchronously: every event accepted
+    // before stop() is drained to disk by the final flush.
+    private val writeReqChannel: Channel<EngineEvent> = Channel(Channel.UNLIMITED)
```
vahidlazio, line 46: before the constructor was empty, now it's unlimited, is this intentional? why did we change it?

fabriziodemaria: Yes, intentional.

**Before (main):** `Channel()` is rendezvous (capacity 0). `emit()`/`flush()` launched a coroutine and called suspending `send()`. `stop()` only cancelled the scope, so events still in the launch queue or mid-send could be lost.

**Now:** `emit()`/`flush()` use synchronous `trySend()` on the caller thread and `stop()` closes the channel + `join()`s `writeJob` (bounded by 2s) so every event accepted before `isStopped = true` is drained to disk.

`trySend()` on a rendezvous channel fails unless the consumer is already blocked on `receive` — with disk I/O and flush-policy work in the writer loop, that would silently drop events (we only log on success). `Channel.UNLIMITED` guarantees the buffer accepts all pre-stop events; covered by `stopDrainsQueuedEventsBeforeUploading` (50 rapid emits → 50 uploads).

Expanded the in-code comment in 8b5910a.

vahidlazio:
`sendChannel` is still a rendezvous channel (capacity 0). When `minBatchSizeFlushPolicy` triggers a flush and an upload is already in flight, the writer blocks on `sendChannel.send(SEND_SIG)` and stops draining `writeReqChannel`. If `stop()` is called during that window, the 2s timeout fires before the writer can consume the remaining buffered events, and `coroutineScope.cancel()` drops them.

Consider making `sendChannel` buffered (e.g. `Channel(Channel.UNLIMITED)`) or switching to `trySend` — the sender already processes all ready files per invocation, so duplicate signals are harmless.

The `stopDrainsQueuedEventsBeforeUploading` test doesn't catch this because the engine is created without `minBatchSizeFlushPolicy`, so the writer never hits `sendChannel.send()`.

vahidlazio:
Here's a failing test that reproduces the issue under production conditions (batch flush policy + slow upload). It currently fails with `expected:<12> but was:<10>` — the writer blocks on the rendezvous `sendChannel.send()` while an upload is in flight, so 2 events never reach disk before the 2s stop timeout.

The fix should make this pass (e.g. buffering `sendChannel` or switching to `trySend`):

```kotlin
@Test
fun stopDrainsAllEventsWhenFlushPolicyBlocksWriter() {
    val slowUploader = SlowEventUploader(uploadDelayMillis = 3_000)
    val batchFlush = object : FlushPolicy {
        private var count = 0
        override fun reset() { count = 0 }
        override fun hit(event: EngineEvent) { count++ }
        override fun shouldFlush(): Boolean = count > 4
    }

    val engine = EventSenderEngineImpl(
        eventStorage = storage,
        clientSecret = "secret",
        uploader = slowUploader,
        flushPolicies = mutableListOf(batchFlush),
        dispatcher = Dispatchers.IO,
        sdkMetadata = SdkMetadata("id", "1.0"),
        debugLogger = null
    )

    repeat(12) { engine.emit("event-$it", mapOf(), mapOf()) }
    Thread.sleep(100)
    engine.stop()

    assertEquals(
        "All 12 events should be written to storage before stop() returns",
        12,
        storage.storedEventCount()
    )
}

private class SlowEventUploader(
    private val uploadDelayMillis: Long
) : EventSenderUploader {
    val uploadedEventNames = mutableListOf<String>()

    override suspend fun upload(events: EventBatchRequest): Boolean {
        kotlinx.coroutines.delay(uploadDelayMillis)
        synchronized(uploadedEventNames) {
            uploadedEventNames.addAll(
                events.events.map { it.eventDefinition.removePrefix("eventDefinitions/") }
            )
        }
        return true
    }
}
```

`RecordingEventStorage` also needs a helper:
```kotlin
fun storedEventCount(): Int = synchronized(this) {
    currentEvents.size + readyEvents.values.sumOf { it.size }
}
```

fabriziodemaria:
Addressed the `sendChannel` blocking issue in `b6048ed`:

- `sendChannel` is now `Channel(CONFLATED)` with `trySend(SEND_SIG)` so the writer loop never suspends on flush signaling while upload holds `uploadMutex`
- Added `stopDrainsAllEventsWhenFlushPolicyBlocksWriter` (your repro with batch flush policy + slow uploader) — asserts all 12 events reach storage before `stop()` returns

The earlier `writeReqChannel` UNLIMITED change only helps when the writer keeps consuming; blocking on rendezvous `sendChannel.send()` was the remaining gap.

Confidence/src/main/java/com/spotify/confidence/EventSenderEngine.kt
```diff
@@ -111,41 +134,100 @@ internal class EventSenderEngineImpl(
...
+        isStopped = true
+        flushIntervalJob?.cancel()
+        // Best effort: drain queued events to disk and attempt one final upload,
+        // bounded so stop() can never block the caller indefinitely. Batches that
+        // don't make it are sealed on disk and retried at next startup.
+        runBlocking(dispatcher) {
+            withTimeoutOrNull(STOP_TIMEOUT_MILLIS) {
+                writeReqChannel.close()
+                writeJob.join()
+                uploadReadyBatches(sealCurrentBatch = true)
```
fabriziodemaria: The 2s timeout also bounds `writeJob.join()`. If draining queued disk writes takes longer, the timeout cancels the join and the subsequent scope cancellation drops events that never reached storage. We should let accepted writes drain fully and bound only the network upload.

Confidence/src/main/java/com/spotify/confidence/PayloadMerger.kt
```diff
@@ -2,11 +2,19 @@ package com.spotify.confidence
...
     override fun invoke(context: ConfidenceStruct, message: ConfidenceStruct): ConfidenceStruct {
-        if (message.containsKey("context")) {
-            throw ConfidenceError.InvalidContextInMessage()
+        return if (message.containsKey("context")) {
+            // An explicit "context" entry in event data overrides the evaluation context for this event.
+            debugLogger?.logMessage(
+                message = "Event data contains a 'context' field: it replaces the evaluation context for this event",
+                isWarning = true
+            )
+            message
```
fabriziodemaria: This returns the caller-owned map directly. Because persistence is asynchronous, mutating a mutable `data` map after `track()` can change the queued payload. Return a snapshot instead.

Confidence/src/main/java/com/spotify/confidence/EventSenderEngine.kt
```diff
@@ -111,41 +134,100 @@ internal class EventSenderEngineImpl(
...
         }
     }
 
     override fun stop() {
+        isStopped = true
+        flushIntervalJob?.cancel()
+        // Best effort: drain queued events to disk and attempt one final upload,
+        // bounded so stop() can never block the caller indefinitely. Batches that
+        // don't make it are sealed on disk and retried at next startup.
+        runBlocking(dispatcher) {
```
fabriziodemaria: `runBlocking(dispatcher)` still blocks the calling thread for up to 2 seconds. `stop()` can be called from Android’s main thread, so final upload should run asynchronously rather than extending teardown latency.

fabriziodemaria (commented):
Bugbot follow-up review: three reliability issues to address before merge.

Confidence/src/main/java/com/spotify/confidence/Confidence.kt
```diff
@@ -386,7 +395,8 @@ object ConfidenceFactory {
         dispatcher: CoroutineDispatcher = Dispatchers.IO,
         loggingLevel: LoggingLevel = LoggingLevel.WARN,
         timeoutMillis: Long = 10000,
-        visitorIdContextKey: String = VISITOR_ID_CONTEXT_KEY
+        visitorIdContextKey: String = VISITOR_ID_CONTEXT_KEY,
+        eventFlushIntervalMillis: Long? = null
```
nicklasl: Could we keep the old `create(...)` JVM signatures around here while adding `eventFlushIntervalMillis`? I guess we will need those bridge overloads so existing compiled consumers do not discover this as a runtime `NoSuchMethodError`.

Confidence/src/main/java/com/spotify/confidence/PayloadMerger.kt
```diff
@@ -2,11 +2,19 @@ package com.spotify.confidence
...
-            throw ConfidenceError.InvalidContextInMessage()
+        return if (message.containsKey("context")) {
+            // An explicit "context" entry in event data overrides the evaluation context for this event.
+            debugLogger?.logMessage(
+                message = "Event data contains a 'context' field: it replaces the evaluation context for this event",
+                isWarning = true
+            )
+            message.toMap()
+        } else {
+            message + mapOf("context" to ConfidenceValue.Struct(context))
```
nicklasl: Could we snapshot the context map before wrapping it in `ConfidenceValue.Struct`? Just want to make sure a caller mutating their per-event context after `track()` cannot rewrite the event while it is still queued.

Provider/src/main/java/com/spotify/confidence/openfeature/ConfidenceFeatureProvider.kt
```diff
@@ -52,7 +52,13 @@ class ConfidenceFeatureProvider private constructor(
         }
     }
 
+    /**
+     * Triggers a best-effort flush of tracked events: delivery is asynchronous
+     * and not guaranteed before process death. Undelivered batches are retried
+     * on the next SDK startup.
+     */
     override fun shutdown() {
+        confidence.flush()
```
nicklasl: Should `shutdown()` call `confidence.stop()` rather than only `flush()` here? I guess we will need the provider lifecycle to cancel the event engine and periodic flush job once OpenFeature shuts it down.

Confidence/src/main/java/com/spotify/confidence/EventSender.kt
```diff
@@ -11,6 +11,16 @@ interface EventSender : Contextual {
         data: ConfidenceFieldsType = mapOf()
     )
 
+    /**
+     * Store a custom event to be tracked with an explicit evaluation context.
+     * @param eventContext evaluation context for this event only; does not mutate session context.
+     */
+    fun track(
```
nicklasl: Could this stay as a `Confidence` overload instead of becoming a new abstract `EventSender` method? Just want to make sure downstream `EventSender` fakes do not have to change for an OpenFeature-specific path.

nicklasl (approved):

PR #253, opened 2026-08-07
feat: improve event delivery reliability for tracked events

fabriziodemaria:
## Summary
- Upload pending `.ready` batches at engine startup (retries events from prior sessions without sealing the current file)
- Flush and upload the current batch in `EventSenderEngine.stop()` before tearing down (called from `Confidence.stop()`)
- Add optional `eventFlushIntervalMillis` to `ConfidenceFactory.create()` for periodic flush (disabled by default)
- Call `confidence.flush()` from provider `shutdown()` as a best-effort lifecycle hook

## Test plan
- [ ] `./gradlew :Confidence:testDebugUnitTest --tests EventSenderEngineReliabilityTest`
- [ ] Verify startup uploads only pending batches and leaves the current file intact
- [ ] Verify `stop()` uploads in-flight events
- [ ] Verify periodic interval triggers upload when configured

Made with [Cursor](https://cursor.com)


fabriziodemaria:
Folded into #252 — track + reliability are now on `openfeature-track-context`.

